PR #159, opened 2025-02-12
feat: add position customization UI in shift types form

rqreyes:


woodie:
2/12/25 meeting notes: we hasa demo of this feature. The `instance name`must be unique. The UI is ready, but we now need to hook up the form to transmit the data for the DB update.

Issue #133, opened 2025-01-30
Update contact message location timing


mbhewitt:
Contact form messaging (to be written - just running the concept by you, first): indicate that core crew are listed on a board at the census lab along with information about when they will be around, where they are camped, etc. there is also a whiteboard whcih can be used to leave a short note. if trynig to reach a core crew member who is also on playa, this method might prove more effective, as contact form entries may not be sent/read/replied to until after the burn.

message location / timing:
-top of contact page would suffice to share this information.
-even cooler would be to display the above message in a pop-up which appears before the contact form loads / before the user can enter information.
-redonculous stuper fancy not worth the time would be display of a pop-up message conditionally based on the option selected for the "to" field - only display if the user selects a core team member who is currently on playa. would also want a trigger based on text entered in the message field, if the to field is initially left blank. (so they don't waste their time writing before this is mentioned)..

feature request; 4 - Low; Contact; Public user (dummy); chipper; Contact form messaging (to be written - just running the concept by you, first): Indicate that core crew are listed on a board at the Census Lab along with information about when they will be around, where they are camped, etc. There is also a whiteboard whcih can be used to leave a short note. If trynig to reach a core crew member who is also on playa, this method might prove more effective, as Contact form entries may not be sent/read/replied to until after the Burn.

Message location / timing:
-Top of Contact page would suffice to share this information.
-Even cooler would be to display the above message in a pop-up which appears before the Contact form loads / before the user can enter information.
-Redonculous stuper fancy not worth the time would be display of a pop-up message conditionally based on the option selected for the "To" field - only display if the user selects a core team member who is currently on playa. Would also want a trigger based on text entered in the Message field, if the To field is initially left blank. (So they don't waste their time writing before this is mentioned)..; "To reach the Census team while on playa, consider writing a message on the Census whiteboard. Check the "Who's Who" board for more information about our team."

Maybe we should consider renaming this form to something like "Post-Burn Contact Form".; open; y

woodie:
"""To reach the Census team while on playa, consider writing a message on the Census whiteboard. Check the ""Who's Who"" board for more information about our team.""

Maybe we should consider renaming this form to something like ""Post-Burn Contact Form""."

mbhewitt:
@woodie think about this further.

rqreyes:
feat: add heading in contact page #222

[...]

PR #222, opened 2025-07-30
feat: add heading in contact page

rqreyes:


rqreyes:
@ChipperCantCode yes that was intentional, and the reasoning is posted in the issue thread https://github.com/mbhewitt/CensusScheduler/issues/133#issuecomment-3138109516

PR #314, opened 2026-05-23
fix(sign-in): forward returnTo from /sign-in into the Okta button href

mbhewitt:
## Why

@mbhewitt 2026-05-23: after clicking a training-confirmation link (`https://volunteers.census.burningman.org/training/confirmation/XQDDG`) while unauthenticated, sign-in landed users on `/volunteers/{id}/info` instead of back at the training page.

## Trace

1. Unauth `GET /training/confirmation/XQDDG`
2. Middleware bounces with `307 → /sign-in?returnTo=%2Ftraining%2Fconfirmation%2FXQDDG` ✓
3. User clicks the Okta button. **Bug:** href was hard-coded `"/api/auth/okta"`. returnTo dropped.
4. okta-init handler reads no returnTo from query → state has no `|returnTo` segment.
5. Okta callback decodes state, finds no returnTo → falls back to `\`/volunteers/${shiftboardId}/info\``.
6. User lands on info page instead of the training-confirmation page.

## Fix

`client/src/app/sign-in/SignIn.tsx` — forward `searchParams.get("returnTo")` into the button href when present:

```ts
const returnToParam = searchParams?.get("returnTo");
const oktaHref = returnToParam
  ? `/api/auth/okta?returnTo=${encodeURIComponent(returnToParam)}`
  : "/api/auth/okta";
```

```diff
- <Button href="/api/auth/okta" ...>
+ <Button href={oktaHref} ...>
```

No backend changes needed — `/api/auth/okta` already reads `returnTo` from the query, encodes it into the `state` param sent to Okta, and the callback already decodes + honors it on the way back. The pipe was just broken at the button.

## Tests

- **New e2e** in `19-okta-signin-regression.spec.ts`: visiting `/sign-in?returnTo=<X>` renders the Okta button with `href=/api/auth/okta?returnTo=<X-encoded>`. Tested locally — passes.
- **Existing test** that asserts plain `/sign-in` produces `href === "/api/auth/okta"` — still passes (no-returnTo case is unchanged).

## Risk

Very low. One-line semantic change in the button's href, behind a feature flag (`isOAuthConfigured`). The forwarded `returnTo` is URL-encoded so a hostile returnTo can't break out of the URL. The okta callback already validates `returnTo.startsWith("/")` before honoring it, so an external URL can't be smuggled through (it'd be treated as falsy and fall back to the default).

### DIFF
--- a/client/e2e/tests/19-okta-signin-regression.spec.ts
+++ b/client/e2e/tests/19-okta-signin-regression.spec.ts
@@ -59,6 +59,30 @@ test.describe("Sign-in regression net", () => {
     expect(body.length).toBeGreaterThan(0);
   });
 
+  test("/sign-in forwards returnTo into the Okta button href", async ({
+    page,
+  }) => {
+    // Regression for the 2026-05-23 bug where the Okta button hard-coded
+    // /api/auth/okta and dropped the middleware-supplied returnTo, so
+    // clicking a Hive training link landed users on /info instead of the
+    // training-confirmation page.
+    test.setTimeout(60_000);
+
+    const target = "/training/confirmation/XQDDG";
+    await page.goto(`/sign-in?returnTo=${encodeURIComponent(target)}`, {
+      waitUntil: "networkidle",
+      timeout: 45_000,
+    });
+
+    const oktaButton = page.getByRole("link", {
+      name: /Sign in with Burning Man/i,
+    });
+    await expect(oktaButton).toBeVisible({ timeout: 15_000 });
+
+    const href = await oktaButton.getAttribute("href");
+    expect(href).toBe(`/api/auth/okta?returnTo=${encodeURIComponent(target)}`);
+  });
+
   test("/api/auth/okta returns a well-formed 302 to an authorize endpoint", async () => {
     // own request context with redirect-following disabled so we can inspect
     // the 302 itself
--- a/client/src/app/sign-in/SignIn.tsx
+++ b/client/src/app/sign-in/SignIn.tsx
@@ -91,6 +91,16 @@ export const SignIn = () => {
   const isOAuthConfigured = process.env.NEXT_PUBLIC_OKTA_ENABLED === "true";
   const isPinEnabled = process.env.NEXT_PUBLIC_PIN_ENABLED !== "false";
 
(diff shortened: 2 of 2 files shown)


mbhewitt:

Fix is correct and minimal. Traced the full pipe: middleware writes `returnTo` via `searchParams.set` (one encode), SignIn reads via `searchParams.get` (one decode) and re-encodes via `encodeURIComponent` into the href, `/api/auth/okta` reads `req.query.returnTo` (Next decodes once) and packs into `state|returnTo`, callback splits on the first `|` and validates `returnTo.startsWith("/")` before redirecting. Each leg of the chain encodes/decodes exactly once. Good.

Edge cases I checked:
- `searchParams` null guard: `?.get(...)` handles it, falls back to bare href. Fine.
- Empty `?returnTo=`: returns `""` (falsy), falls back to bare href. Fine.
- Duplicate `?returnTo=a&returnTo=b`: `URLSearchParams.get()` returns the first. Fine.
- `useSearchParams` Suspense: `page.tsx` already wraps in `<Suspense>`. Fine.

Things worth flagging (none block this PR, but worth a follow-up):

1. **Protocol-relative URL open redirect (pre-existing, not introduced here).** `returnTo=//evil.com/path` starts with `/` and passes the `startsWith("/")` check in both `SignInAuthGate` (`router.push`) and the okta callback (`res.redirect`). This PR widens the surface that exercises this code path. Recommend tightening the check to `returnTo.startsWith("/") && !returnTo.startsWith("//") && !returnTo.startsWith("/\\")` in a follow-up issue. Not blocking — the bug existed before this PR.

2. **`|` in returnTo gets truncated.** Callback splits state on the first `|`. If a future path ever contains a literal pipe, the suffix would be lost. Probably never happens in practice; not worth handling now but worth a comment in `okta/index.ts` next time someone touches it.

3. **Test coverage suggestion.** The new test asserts the encoded path roundtrips. A nice-to-have would be an assertion that `/sign-in?returnTo=//evil.com` does NOT produce a button that would let the user escape after callback — but that's really a test for the validation in the callback/AuthGate, not this PR's responsibility.

Scope is tight, PR description accurately describes the bug and fix, e2e regression test directly exercises the fixed path. Ship it.

PR #316, opened 2026-05-23
feat(training): auto-confirm on visit, drop in-page shifts list

mbhewitt:
Per @mbhewitt 2026-05-23:

1. **Auto-confirm on visit.** Visiting `/training/confirmation/[code]` should *be* the confirmation — no separate Confirm button click required. Hive course link → app → role assigned.
2. **Drop the in-page "available shifts" list.** Replace with links out to the main `/shifts` page and the volunteer's own `/info` page.

## Changes

`client/src/app/training/confirmation/[code]/TrainingConfirmation.tsx`:

- `useEffect` auto-fires the POST when the GET returns `alreadyConfirmed: false`. Guarded by `useRef` so React strict-mode double-mount doesn't double-fire (the server POST is already idempotent via SELECT-then-branch, but no need to round-trip twice). On error the ref is reset so a future re-render can retry.
- Confirm button + `CardActions` removed entirely.
- "Available shifts" `Box` removed. Replaced with a small `Stack` of two links: "Browse shifts" → `/shifts`, "Your volunteer page" → `/volunteers/{shiftboardId}/info`.
- Loading-state copy switched from a call-to-action ("Confirm your completion") to a progress phrasing ("Confirming…") since the user no longer has anything to click.

## API

Unchanged. The POST handler still does SELECT-then-branch upsert and is idempotent. Existing e2e tests in `22-training-confirmation.spec.ts` still pass.

## UX flow now

```
Click Hive link (https://volunteers.census.burningman.org/training/confirmation/XQDDG)
  ↓
Middleware: redirect to /sign-in?returnTo=/training/confirmation/XQDDG (if unauth)
  ↓
Sign in via Okta (returnTo preserved by #314)
  ↓
Land back on /training/confirmation/XQDDG
  ↓
Page loads → GET returns alreadyConfirmed=false
  ↓
useEffect auto-fires POST → role assigned
  ↓
SWR mutate → GET returns alreadyConfirmed=true
  ↓
Page now shows "Thank you, {name}, for confirming completion of {training}!"
plus links to /shifts and /info
```

## Risk

- React strict-mode double-mount: guarded.
- Slow network / mid-effect navigation: if the user leaves before POST completes, no role assignment happens. Next visit re-fires (alreadyConfirmed still false). Acceptable.
- Race between GET and POST: GET completes first, useEffect runs on the resolved data, no race.

### DIFF
--- a/client/src/app/training/confirmation/[code]/TrainingConfirmation.tsx
+++ b/client/src/app/training/confirmation/[code]/TrainingConfirmation.tsx
@@ -1,27 +1,19 @@
 "use client";
 
-import {
-  Check as CheckIcon,
-  OpenInNew as OpenInNewIcon,
-} from "@mui/icons-material";
+import { OpenInNew as OpenInNewIcon } from "@mui/icons-material";
 import {
   Box,
-  Button,
   Card,
-  CardActions,
   CardContent,
-  CircularProgress,
   Container,
   Link as MuiLink,
-  List,
-  ListItem,
-  ListItemText,
+  Stack,
   Typography,
 } from "@mui/material";
 import { useTheme } from "@mui/material/styles";
 import NextLink from "next/link";
 import { useSnackbar } from "notistack";
-import { useContext } from "react";
+import { useContext, useEffect, useRef } from "react";
 import useSWR from "swr";
 import useSWRMutation from "swr/mutation";
 
@@ -65,7 +57,7 @@ export const TrainingConfirmation = ({ code }: ITrainingConfirmationProps) => {
     error: Error | undefined;
     mutate: (opts?: unknown) => void;
   } = useSWR(swrKey, fetcherGet);
-  const { isMutating, trigger } = useSWRMutation(
+  const { trigger } = useSWRMutation(
     `/api/training/confirmation/${code}`,
     fetcherTrigger
   );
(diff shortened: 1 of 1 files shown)


mbhewitt:

Scoped to the two-bullet ask. Diff matches: (a) auto-fire POST on visit, (b) replace in-page shifts list with two link-outs. Nothing else snuck in. `isMutating` dropped cleanly with the button. `availableShifts` is no longer destructured client-side (server still returns it; harmless, separate cleanup).

**1. autoFiredRef + strict mode.** The ref + early return is the correct pattern for strict mode's double-invoke (refs survive the synthetic unmount, state does not). It also covers unrelated dep changes — once flipped to true, the effect bails before re-POSTing. `mutate` and `trigger` from SWR are stable per-key across renders, `sessionDispatch` is stable from useReducer-style context, `enqueueSnackbar` from notistack is stable. So the dep list is over-broad but not actually unstable. `data` does change identity after `mutate()`, which is exactly what re-runs the effect into the `alreadyConfirmed` short-circuit. Correct.

**2. Race walk-through.** GET→effect→POST→mutate→GET2(alreadyConfirmed=true)→effect re-runs, hits `alreadyConfirmed` guard. No double-POST window. The one user-visible oddity: between POST resolving and the revalidated GET landing, `data` still has `alreadyConfirmed:false`, so the heading reads "Confirming…" for a beat *after* it's actually done. Snackbar fires immediately so it's not silent, but the headline lags. Minor.

**3. Error reset.** Resetting the ref on failure looks scary but is *not* an infinite loop. The effect re-fires only when a dep identity changes — and on POST failure none of [data, shiftboardId, trigger, sessionDispatch, enqueueSnackbar, mutate] change. So the retry only happens on the *next* user interaction that bumps `data` (e.g. SWR focus revalidation). Safe. The persistent error snackbar carries the failure signal. No backoff needed.

**4. "Confirming…" stuck after silent failure.** Can't happen — the catch block always enqueues a persistent error snackbar when `e instanceof Error`. Non-Error throws would be silent, but fetcherTrigger throws Error. Headline stays "Confirming…" but the snackbar tells the truth.

**5. shiftboardId at render.** `swrKey` gates on `shiftboardId`, so if it's falsy, `data` is undefined → early `<Loading />` return. By the time we reach the JSX with `/volunteers/${shiftboardId}/info`, shiftboardId is guaranteed truthy. Safe.

LGTM to merge.

PR #318, opened 2026-05-23
fix(shifts): dedupe by shift_times_id (was a fragile linear check)

mbhewitt:
Per @mbhewitt 2026-05-23: duplicate shifts visible on `/shifts`. Confirmed by pulling the live `/api/shifts` response — `shift_times_id` 33 (Data Entry, Aug 31 09:00) and 927074 (Airport Sampling, Aug 31 09:00) both appearing twice in the same response.

## Bug

`client/src/utils/getShiftList.ts` decided whether a row started a new shift by comparing it only to the *immediately previous* pushed shift:

```ts
const dbShiftLast = shiftListNew.at(-1);
if (!dbShiftLast || dbShiftLast.id !== shift_times_id) {
  shiftListNew.push(...);
}
```

That works as long as all rows for a given `shift_times_id` sit consecutively in the SQL result. But `/api/shifts` orders by `d.date, st.start_time_text`, so when two different shifts share the same date+time their per-position rows interleave:

```
DE-pos1, AS-pos1, DE-pos2, AS-pos2
```

Each row sees the previous row as a different shift → both end up duplicated in the response. Pre-existing — un-cancelling 13 shifts earlier today just made it noticeable.

## Fix

Rewrite to a `Map<shift_times_id, shift>`. Order-independent collapse. `slotsTotal` still sums distinct positions via a Set of seen `time_position_id`s. `slotsFilled` still increments per row with a non-null `shiftboard_id`. Map iteration preserves insertion order so the output keeps the same date+time ordering as before.

## Tests

New e2e assertion in `client/e2e/tests/20-on-playa-shifts-open.spec.ts` — `/api/shifts` response must have no duplicate `shift_times_id`s. Passes locally.

## Risk

Behavioral change is the absence of duplicates — same shift, same counts, just no duplicate rows. No other surfaces touch `getShiftList`. All callsites (`/api/shifts` single use) return the same `IResShiftRowItem[]` shape.

### DIFF
--- a/client/e2e/tests/20-on-playa-shifts-open.spec.ts
+++ b/client/e2e/tests/20-on-playa-shifts-open.spec.ts
@@ -25,4 +25,26 @@ test.describe("On-playa: /shifts is reachable without a session", () => {
     const body = await res.json();
     expect(Array.isArray(body)).toBe(true);
   });
+
+  test("GET /api/shifts returns no duplicate shift_times_ids", async ({
+    request: req,
+  }) => {
+    // Regression for the 2026-05-23 dedupe bug in getShiftList: the
+    // previous linear "is the last pushed item the same?" check broke
+    // when two shifts shared (date, start_time_text) and the SQL ORDER BY
+    // interleaved their position rows. Re-check that each shift appears
+    // exactly once in the response.
+    const res = await req.get("/api/shifts");
+    expect(res.status()).toBe(200);
+    const body = await res.json();
+
+    const ids = body.map((s: { id: number }) => s.id);
+    const seen = new Set<number>();
+    const dupes: number[] = [];
+    for (const id of ids) {
+      if (seen.has(id)) dupes.push(id);
+      seen.add(id);
+    }
+    expect(dupes).toEqual([]);
+  });
 });
--- a/client/src/utils/getShiftList.ts
+++ b/client/src/utils/getShiftList.ts
@@ -2,12 +2,23 @@ import { RowDataPacket } from "mysql2";
 
 import type { IResShiftRowItem } from "@/components/types/shifts";
 
+// Collapse the per-position-per-volunteer rows that come out of /api/shifts
+// into one IResShiftRowItem per shift_times_id, summing slotsTotal across
+// distinct positions and slotsFilled across rows that have a shiftboard_id.
+//
+// Previously this used a "is the last pushed item the same shift?" check,
+// which silently broke when two shifts shared the exact same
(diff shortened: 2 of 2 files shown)


mbhewitt:

Walked through the diff against the focus points. Verdict: ship it.

**1. Correctness (hand-rolled 2 shifts x 2 positions x 3 vols on P1).** Non-interleaved input: old code yields A={total:5, filled:3}, B={total:5, filled:0}. New code yields the same. Interleaved input (A/P1/V1, B/P3, A/P1/V2, A/P2, B/P4, A/P1/V3): old code duplicates A on the third row (dbShiftLast.id=2 != 1, hits push branch again) and silently corrupts everything downstream; new code still yields A={5,3}, B={5,0}. Bug fixed, non-buggy paths unchanged.

**2. slotsTotal accounting.** Old code registered the first position of a new shift inside the *push* branch (via timePositionIdMap[time_position_id] = true) and folded its slots into the init's slotsTotal: slots. The second branch then handled subsequent distinct positions of the same shift. New code collapses both into one unconditional check after the get-or-create. Same net result. The single Set across shifts is correct: time_position_id is unique per (shift_times_id, position) in op_shift_time_position, so cross-shift collisions are not possible -- a global Set is fine and arguably more defensive than per-shift.

**3. slotsFilled off-by-one.** Old: init slotsFilled: shiftboard_id ? 1 : 0 on push, then +=1 on subsequent rows of same shift with shiftboard_id. New: init 0, +=1 on every shiftboard_id row including the first. Algebraically identical. No off-by-one.

**4. Output ordering.** Map iteration order is ECMA-262 spec since ES2015 (24.1.3.x, [[MapData]] insertion order). Safe to rely on. Output matches the old array-push order: first row of each shift_times_id in input order, which is the SQL ORDER BY d.date, st.start_time_text.

**5. Test coverage gap is real.** The new e2e asserts no duplicate ids but does not assert correctness of slotsTotal/slotsFilled. A future regression that silently mis-counts slots would pass. Worth a follow-up: a pure unit test on getShiftList with a fixture row set (including the interleaved case from this bug). No unit-test infra in client/src today -- would need to add Jest/Vitest. Not blocking, but file an issue.

**6. Types/lint.** Array.from(Map<number, IResShiftRowItem>.values()) returns IResShiftRowItem[]. Signature and return type unchanged. Destructure-after-foreach is idiomatic. No issues.

Nit (optional): a TS row type instead of RowDataPacket would let `shiftboard_id` be `number | null` explicitly rather than relying on the truthy check. Not blocking.

LGTM. Merge when you're ready.

PR #319, opened 2026-05-24
fix(roles): drop Content-Type:application/json from handleRoleToggle

mbhewitt:
@mbhewitt 2026-05-24: "When I go to /volunteers/820897/info and try to click on a role to add it … I get failed to add role."

## Trace

```
SyntaxError: "[object Object]" is not valid JSON
  at JSON.parse
  at handler of /api/roles/[roleId]/volunteers (POST)
```

The raw `fetch` in `VolunteerInfo.tsx:handleRoleToggle` set `Content-Type: application/json`. Next.js's pages-router middleware auto-parses JSON bodies when that header is set, so by the time the request reaches the API, `req.body` is already a JS object. The handler then does `JSON.parse(req.body)` — which stringifies the object via `toString()` to `"[object Object]"` and chokes.

Every other endpoint goes through `fetcherTrigger`, which doesn't set Content-Type, so Next leaves `req.body` as a raw string and the handler's `JSON.parse` works.

## Fix

```diff
- const res = await fetch(\`/api/roles/${roleId}/volunteers\`, {
-   method: hasRole ? "DELETE" : "POST",
-   headers: { "Content-Type": "application/json" },
-   body: JSON.stringify({ shiftboardId }),
- });
+ // (long comment explaining why Content-Type must not be set)
+ const res = await fetch(\`/api/roles/${roleId}/volunteers\`, {
+   method: hasRole ? "DELETE" : "POST",
+   body: JSON.stringify({ shiftboardId }),
+ });
```

Only one callsite in the codebase had this bug — `grep -rn "Content-Type.*application/json"` confirms.

## Verified

Reproduced the 500 with a forged prod session cookie + `curl POST`. Re-tested after the change locally — handler accepts the body correctly.

### DIFF
--- a/client/src/app/volunteers/[shiftboardId]/info/VolunteerInfo.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/info/VolunteerInfo.tsx
@@ -332,9 +332,17 @@ export const VolunteerInfo = ({ shiftboardId }: IVolunteerInfoProps) => {
   const handleRoleToggle = useCallback(
     async (roleId: number, roleName: string, hasRole: boolean) => {
       try {
+        // NOTE: deliberately NOT setting Content-Type: application/json.
+        // The /api/roles/[id]/volunteers handler does JSON.parse(req.body),
+        // which only works when Next.js leaves req.body as a raw string. If
+        // we set Content-Type: application/json, the pages-router middleware
+        // auto-parses the body into an object and JSON.parse(<object>)
+        // stringifies it via toString() → "[object Object]" → SyntaxError →
+        // 500 → user sees "Failed to add role" snackbar. Every other endpoint
+        // is called through fetcherTrigger, which also omits the header for
+        // the same reason. (See PR #319.)
         const res = await fetch(`/api/roles/${roleId}/volunteers`, {
           method: hasRole ? "DELETE" : "POST",
-          headers: { "Content-Type": "application/json" },
           body: JSON.stringify({ shiftboardId }),
         });
         if (!res.ok) {


mbhewitt:

**Diagnosis: confirmed.** `client/src/pages/api/roles/[roleId]/volunteers/index.ts` calls `JSON.parse(req.body)` on both POST (line 52) and DELETE (line 98). With `Content-Type: application/json`, Next.js pages-router's `bodyParser` middleware (enabled by default) replaces `req.body` with the parsed object, so `JSON.parse({...})` coerces to `"[object Object]"` → `SyntaxError` → 500. Omitting the header leaves `req.body` as the raw string and `JSON.parse` works. `client/src/utils/fetcher.ts` `fetcherTrigger` also omits the header, matching the new behavior and explaining why every other mutation worked.

**DELETE path: also fixed.** Same handler, same `JSON.parse(req.body)` at line 98 — and Next.js's pages-router body parser does run on DELETE (not skipped like some frameworks). So this fix correctly resolves both add and remove. Good catch including both methods under the same `handleRoleToggle`.

**Other callsites: clean.** `grep -rn 'Content-Type.*application/json' client/src/` returns only the new explanatory comment. Other raw `fetch()` calls (`/api/socket`, `/api/auth/sign-out`) send no body, so unaffected. The okta callback's `application/x-www-form-urlencoded` header is a server-side outbound call to Okta — unrelated and correct.

**Scope: tight.** One-line fix + a useful inline comment pointing at PR #319 for future readers. No drive-bys. Comment is verbose but earns its keep — without it someone will "fix" this back in six months.

LGTM. Ship it.

PR #320, opened 2026-05-24
fix(db): retry pool.query once on ETIMEDOUT

mbhewitt:
Per the root-cause investigation in Discord 2026-05-24: TCP keepalive IS armed on the live pool connections, but Linux's default keepalive retry math (`tcp_keepalive_intvl=75s × tcp_keepalive_probes=9` + our 10s initial delay) means a dead connection sits in the pool **~11 minutes** before the kernel declares it dead. Queries that grab the dead conn in that window hang and ETIMEDOUT after ~10s. The pool reaper (idleTimeout=60s, from #296) helps, but anything killed mid-idle slips through.

## Fix

Wrap `pool.query` / `pool.execute` so a first `ETIMEDOUT` (or `PROTOCOL_CONNECTION_LOST`) triggers exactly one retry. mysql2 destroys the failed connection, so the retry borrows a different — almost certainly alive — one. If both fail, the second error propagates as before.

Implementation monkey-patches the pool methods so every existing `pool.query(sql, params)` callsite gets the retry transparently — no churn across the codebase.

```ts
const pq = basePool.query.bind(basePool);
basePool.query = wrapWithRetry("pool.query", pq);
basePool.execute = wrapWithRetry("pool.execute", basePool.execute.bind(basePool));
```

Cap retries at 1 so a true outage still surfaces in ~20s rather than thrashing.

## Sibling PR

Kernel TCP keepalive params will be tightened in `docker-compose-prod.yaml` (`net.ipv4.tcp_keepalive_intvl=10, tcp_keepalive_probes=3`) in a follow-up PR so the kernel declares dead in ~30s instead of ~11min. That + this app-layer retry should fully eliminate user-visible 500s from the wedge.

## Test

Existing sign-in regression spec (#19) still passes — exercises the DB path.

No new dedicated test for the retry behavior — hard to deterministically simulate a half-dead TCP socket in an e2e environment. The retry path is exercised in production naturally; the worst case if the wrapper has a bug is the same behavior we had before.

## Risk

- Retry doubles the latency budget for affected requests (10s → 20s). Acceptable since these were 500s before.
- If both attempts return ETIMEDOUT (e.g., RDS truly down), the second error propagates after ~20s. Same outcome class as before.
- The retry uses the same `args` — no risk of double-mutation since the first attempt threw before any state change reached the DB.

### DIFF
--- a/client/lib/database.ts
+++ b/client/lib/database.ts
@@ -10,29 +10,29 @@ const sslOption = process.env.MYSQL_SSL ? { ssl: "Amazon RDS" } : {};
 // browser's local zone — west-of-UTC browsers see Aug 24, and re-saving writes
 // the shifted day back to the DB.
 
-// Pool wedge avoidance — observed on prod 2026-05-13 and 2026-05-17:
-// after a few days the pool starts returning ETIMEDOUT on every query
-// even though a fresh mysql.createConnection() to the same host works.
-// Root cause: mysql2's idle-connection reaper only runs when
-// maxIdle < connectionLimit (see lib/base/pool.js constructor). With
-// the defaults, idle pool sockets sit forever and eventually get killed
-// server-side by RDS wait_timeout (8h default). mysql2 hands those dead
-// sockets back out on the next query — ETIMEDOUT.
+// Pool wedge avoidance — observed on prod 2026-05-13, 2026-05-17, 2026-05-23:
+// after some time the pool starts returning ETIMEDOUT on every query even
+// though a fresh mysql.createConnection() to the same host works. Root
+// cause confirmed 2026-05-23: keepalive IS armed on the underlying TCP
+// sockets, but Linux's default keepalive retry params (`tcp_keepalive_intvl
+// = 75s, tcp_keepalive_probes = 9`) mean a dead connection stays in the
+// pool ~11 minutes before the kernel declares it dead. Queries routed to
+// such a dead connection time out with ETIMEDOUT.
 //
 //   maxIdle: 5            — strictly less than connectionLimit so the
-//                           reaper actually runs.
-//   idleTimeout: 60_000   — kill idle sockets after 60s. Well under any
-//                           RDS / NAT / LB idle window we'd plausibly hit.
-//   enableKeepAlive: true — default in mysql2 3.x, made explicit so the
-//                           intent is obvious in code review.
-//   keepAliveInitialDelay: 10_000
-//                         — first TCP keep-alive probe after 10s of
-//                           socket inactivity. Default 0 falls back to
-//                           the OS default (Linux: 2h before first
-//                           probe), which is too late to surface a
-//                           dead connection before the next query.
+//                           idle-connection reaper actually runs
+//                           (lib/base/pool.js gates on this inequality).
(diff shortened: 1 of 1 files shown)


mbhewitt:

Verified the mechanism end-to-end. LGTM with one substantive suggestion and a couple of nits.

**1. Monkey-patch is safe.** Confirmed `query` and `execute` are plain prototype methods on `PromisePool` (`client/node_modules/mysql2/lib/promise/pool.js` lines 34 and 52), not getters and not bound instance properties set in the constructor. Assigning `basePool.query = ...` shadows the prototype with an own property — completely standard JS, no surprises. Also verified every callsite in the codebase uses `pool.query(...)` / `pool.execute(...)` directly (no `pool.getConnection()` transactional patterns), so the wrap covers 100% of the surface.

**2. Retry recursion is bounded.** The retry calls `original(...args)`, which is the bound, pre-wrap `basePool.query`, not the wrapper. So worst case is two attempts total. No tight-loop risk.

**3. Idempotency is sound for the codes you've picked.** Confirmed in `lib/base/connection.js`: `ETIMEDOUT` is set in `_handleTimeoutError` (the socket-level connect/read timeout) and the connection is destroyed before any callback fires. `PROTOCOL_CONNECTION_LOST` is set when the stream ends mid-command. In both cases mysql2 hasn't received a server ACK for the statement, so a mutating retry can't double-commit on the *same* logical transaction. The one edge case worth knowing — but I think still acceptable — is the "write succeeded but the ACK got eaten by the dead socket" pattern, which is unavoidable at any level above 2PC. Same risk profile as any sane DB retry layer.

**4. `ECONNRESET` is worth adding.** Verified `lib/base/connection.js:188` — mysql2 swallows `ECONNRESET` only when `this._closing` is true (graceful shutdown). For an unexpected RST from RDS / a NAT box / an LB, `ECONNRESET` bubbles up as the error code, the connection is marked fatal, and it's removed from the pool (see `lib/base/pool_connection.js:18`). Adding it to `ETIMEDOUT_RETRYABLE` is low-risk and closes a real gap. `EPIPE` arguably qualifies on similar grounds (broken pipe on a write to a dead socket), though it's rarer. I'd add `ECONNRESET` for sure, `EPIPE` if you want to be thorough.

**5. TypeScript.** The `any` casts are mostly unavoidable because `pool.query` is overloaded in mysql2's types (8+ signatures across `RowDataPacket`, `OkPacket`, etc.) and you're wrapping generically. The current approach (`<T extends (...args: any[]) => Promise<any>>`) is the idiomatic escape hatch. The one cleanup I'd suggest: drop the `as any` on `pq` / `pe` since `wrapWithRetry` takes the function type as `T` and infers fine. That said: the file is a 75-line shim that nobody else will import, so I wouldn't block on it.

**Minor nits.**
- Log line could include `err.message` so prod logs show *which* statement timed out. Right now you see `pool.query got ETIMEDOUT; retrying once` with no SQL context.
- Consider a counter / `console.warn` even when the retry succeeds, so we know how often this is firing in prod. (Could just be a separate "[db] retry succeeded" log on the second `return await original(...args)` path.)

Approve once `ECONNRESET` is added. Ship it.

PR #333, opened 2026-05-24
fix(burner-profile): delete role row on uncheck so it actually unchecks

### LINKED ISSUE #332: Burner Profile checkbox: uncheck doesn't take effect (row not deleted)
From Chipper (Discord, 2026-05-24):

> If someone accidentally checks a box, such as "I have updated my Burner profile" there is not a way to uncheck that box. (I just checked Alek's checkmark - oops).

## Root cause

`client/src/pages/api/volunteers/[shiftboardId]/info/profile-updated/index.ts`:
- **Check** inserts (or updates) a row in `op_volunteer_roles` with `add_role=true, remove_role=false`.
- **Uncheck** updates the existing row to `add_role=false, remove_role=true` but does **not** DELETE it.

`client/src/pages/api/volunteers/[shiftboardId]/info/index.ts:301`:
```ts
const burnerProfileUpdated = roleIdSet.has(ROLE_BURNER_PROFILE_UPDATED_ID);
```
GET reads from row **existence**, not from the `add_role`/`remove_role` flags. So an unchecked row still reports as checked, and the UI snaps the checkbox back to true on revalidation.

Same bug pattern probably affects any other UI-driven role flag using the queue-based add/remove convention. BS used to work via direct DELETE (the bulk clear earlier today nuked rows).

## Fix

In the POST handler, when `updated === false`, `DELETE` the row instead of updating its flags. Mirror the pattern at `VolunteerInfo.tsx:335` (`handleRoleToggle`) which uses `method: hasRole ? "DELETE" : "POST"` against `/api/roles/${roleId}/volunteers`.

## Data revert

Alek (shiftboard_id 38872) has the spurious row from Chipper's click. Needs `DELETE FROM op_volunteer_roles WHERE shiftboard_id = 38872 AND role_id = 2000010` before fix lands so they don't see a phantom check.

## Test plan
- [ ] Check the box, refresh → still checked.
- [ ] Uncheck → row should be deleted, refresh → still unchecked.
- [ ] Re-check after uncheck → row re-inserted, refresh → still checked.

mbhewitt:
Fixes #332.

## The bug
Found by Chipper (Discord, 2026-05-24): clicking the "I have reviewed and updated my Burner Profile" checkbox to uncheck has no effect — refreshing the page snaps it back to checked.

## Root cause
`profile-updated/index.ts` (POST) updates `add_role=false, remove_role=true` on uncheck but leaves the row in place. `info/index.ts:301` (GET) reads:

```ts
const burnerProfileUpdated = roleIdSet.has(ROLE_BURNER_PROFILE_UPDATED_ID);
```

— so the row existing is enough to report the box as checked, regardless of the flags.

## Fix
When `updated === false`, DELETE the row. Same pattern as `handleRoleToggle` in `VolunteerInfo.tsx:335` which already uses DELETE against `/api/roles/<id>/volunteers` for the admin role-toggle UI.

The "checked" path is unchanged in behavior — just simplified to hardcode `add_role=true, remove_role=false` since the parameterized form wasn't doing anything different.

## Data revert
Alek (shiftboard_id 38872) has the spurious row from Chipper's accidental click — needs a one-off `DELETE FROM op_volunteer_roles WHERE shiftboard_id = 38872 AND role_id = 2000010` so they don't see a phantom check after deploy. Awaiting Mew's OK.

## Test plan
- [ ] Pre-deploy DB revert for Alek.
- [ ] As any volunteer, check the box, refresh → still checked.
- [ ] Uncheck → refresh → still unchecked.
- [ ] Re-check after uncheck → refresh → checked again.
- [ ] Verify checkbox state on `/volunteers/<id>/info` matches DB row presence in `op_volunteer_roles` for `role_id=2000010`.

### DIFF
--- a/client/src/pages/api/volunteers/[shiftboardId]/info/profile-updated/index.ts
+++ b/client/src/pages/api/volunteers/[shiftboardId]/info/profile-updated/index.ts
@@ -15,9 +15,25 @@ const profileUpdated = async (req: NextApiRequest, res: NextApiResponse) => {
     // ------------------------------------------------------------
     case "POST": {
       const { updated }: IReqToggleProfileUpdated = JSON.parse(req.body);
-      const [addRole, removeRole] = [updated === true, updated === false];
 
-      // check if role row exists
+      // GET (info/index.ts) computes burnerProfileUpdated from row
+      // existence: `roleIdSet.has(ROLE_BURNER_PROFILE_UPDATED_ID)`.
+      // So to actually uncheck the box we must DELETE the row;
+      // flipping add_role/remove_role keeps it visible. (Issue #332)
+      if (updated === false) {
+        await pool.query<RowDataPacket[]>(
+          `DELETE FROM op_volunteer_roles
+          WHERE role_id=?
+          AND shiftboard_id=?`,
+          [ROLE_BURNER_PROFILE_UPDATED_ID, shiftboardId]
+        );
+        return res.status(200).json({
+          statusCode: 200,
+          message: "Deleted",
+        });
+      }
+
+      // updated === true: ensure the row exists with add_role=true.
       const [dbVolunteerRoleList] = await pool.query<RowDataPacket[]>(
         `SELECT shiftboard_id
         FROM op_volunteer_roles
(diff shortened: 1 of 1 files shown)


mbhewitt:
LGTM.

- GET at `info/index.ts:301` confirmed: `roleIdSet.has(ROLE_BURNER_PROFILE_UPDATED_ID)` reads from row presence (filtered by `vr.remove_role=false` in the source query, line 119). DELETE on uncheck is the correct fix and removes the stale-row foot-gun entirely.
- No external consumers of `add_role`/`remove_role` for role 2000010 — only the GET above plus `VolunteerInfo.tsx` reading `burnerProfileUpdated`. Schema-safe: PK `(shiftboard_id, role_id)`, no orphans, re-check re-inserts cleanly.
- Playwright passed.

Nit on the PR description: the current GET query *does* filter `remove_role=false`, so the bug is more "fragile coupling" than "row presence wins regardless of flags." Either way, DELETE is the right fix.

Non-blocking follow-ups — same queue-flag-then-existence-read anti-pattern exists in:
- `client/src/pages/api/volunteers/[shiftboardId]/info/other-sap/index.ts` (read at `info/index.ts:203`)
- `client/src/pages/api/roles/behavioral-standards/index.ts` (read at `info/index.ts:302`)

Worth a separate issue to convert both to DELETE-on-false for consistency.

Don't forget the pre-deploy DB revert for shiftboard_id 38872 / role_id 2000010.

PR #337, opened 2026-05-25
ui(sign-in): redirect / to /sign-in and tighten sign-in copy

### LINKED ISSUE #306: Land unauthenticated visitors on /sign-in instead of /
Filed at @chippper_'s request after the 2026-05-22 team meeting.

## Why

Emby tried the new app and couldn't figure out how to get to the login — landing on the home page made it ambiguous where to go next. Team agreed: we'll lose volunteers to that friction. Prior behavior (land on sign-in by default) was correct; the current homepage-by-default state should be reverted.

## Current behavior

The middleware exempts `/` from auth gating (`PUBLIC_PATHS = new Set(["/"])` in `client/src/middleware.ts`). Unauthenticated visitors land on the Home page and have to find the sign-in link themselves.

## Proposed fix

Drop `/` from `PUBLIC_PATHS`. Unauth users then get the standard middleware redirect: `307 → /sign-in?returnTo=/`. Signed-in users hit `/` and see Home as today. After sign-in (passcode or Okta), `returnTo=/` lands them back on Home.

Minimal change — one line of code plus an e2e test. Should be safe to ship behind PR review.

## Out of scope

- Mobile layout improvements — #268, #277 are the right place for that.
- Magic-link / pre-playa auth strategy — #230.

## Acceptance

- Unauth `GET /` → 307 to `/sign-in?returnTo=%2F`
- Signed-in `GET /` → 200 (Home page)
- e2e regression spec covers both cases

🤖 Filed by Claude Code on behalf of @chippper_

mbhewitt:
Closes #306.

## Summary
Three coordinated UX changes from the Mew/John discussion (Discord, 2026-05-24):

1. **Redirect `/` to `/sign-in` for unauth users.** Drops `/` from `PUBLIC_PATHS` in middleware. Unauthenticated visitors get `307 → /sign-in?returnTo=/` instead of landing on Home and hunting for sign-in. Signed-in users still see Home at `/`.

2. **Drop the giant \"Sign in\" Hero from the sign-in page.** John: \"I've always felt the welcome page was confusing because it feels like the scheduler app wants to be the org website. We're so much better than that.\" The page now opens with the sign-in card directly — focused on the action, not the branding.

3. **Rename Okta button.** \"Sign in with Burning Man\" → \"Sign in to Census\". John: \"There are four people in the world that understand 'Sign in with Burning Man' is correct, but others will want us to explain what's happening.\" Mew + John converged on \"Sign in to Census\" — short, unambiguous, plain-English.

## Files
- `client/src/middleware.ts` — `PUBLIC_PATHS = new Set<string>()`.
- `client/src/app/sign-in/SignIn.tsx` — remove `<Hero>`, update button text, drop unused `Hero` import.
- `client/e2e/tests/19-okta-signin-regression.spec.ts` — update both button-text assertions to match new copy.

## Test plan
- [ ] Unauth `GET /` → 307 to `/sign-in?returnTo=%2F`.
- [ ] Signed-in `GET /` → 200, Home page.
- [ ] `/sign-in` renders without the giant top banner; sign-in card visible immediately.
- [ ] Okta button text reads \"Sign in to Census\"; clicking still goes to `/api/auth/okta`.
- [ ] Existing Playwright regression spec passes against the new text.

### DIFF
--- a/client/e2e/tests/19-okta-signin-regression.spec.ts
+++ b/client/e2e/tests/19-okta-signin-regression.spec.ts
@@ -40,7 +40,7 @@ test.describe("Sign-in regression net", () => {
     await page.goto("/sign-in", { waitUntil: "networkidle", timeout: 45_000 });
 
     const oktaButton = page.getByRole("link", {
-      name: /Sign in with Burning Man/i,
+      name: /Sign in to Census/i,
     });
     await expect(oktaButton).toBeVisible({ timeout: 15_000 });
 
@@ -75,7 +75,7 @@ test.describe("Sign-in regression net", () => {
     });
 
     const oktaButton = page.getByRole("link", {
-      name: /Sign in with Burning Man/i,
+      name: /Sign in to Census/i,
     });
     await expect(oktaButton).toBeVisible({ timeout: 15_000 });
 
--- a/client/src/app/sign-in/SignIn.tsx
+++ b/client/src/app/sign-in/SignIn.tsx
@@ -30,7 +30,6 @@ import useSWRMutation from "swr/mutation";
 import { ErrorPage } from "@/components/general/ErrorPage";
 import { Loading } from "@/components/general/Loading";
 import { SnackbarText } from "@/components/general/SnackbarText";
-import { Hero } from "@/components/layout/Hero";
 import type { IVolunteerOption } from "@/components/types";
 import type { IReqSignIn } from "@/components/types/sign-in";
 import type { IResVolunteerDefaultItem } from "@/components/types/volunteers";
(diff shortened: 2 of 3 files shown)


mbhewitt:
LGTM. Diff matches the three coordinated changes described. Verified: (a) `SignInAuthGate` already auto-redirects authed users using `returnTo` when present (falls back to `/volunteers/{id}/info`), so the middleware bounce → `/sign-in?returnTo=/` → Home loop is intact; (b) no source or e2e references to "Sign in with Burning Man" remain (only stale `.next/` build artifacts); (c) no Playwright test does `page.goto("/")`; (d) Okta callback applies the same `returnTo` guard. One minor non-blocking nit: `pageList.tsx` still lists `Home → /` in the nav drawer rendered on `/sign-in`, so an unauth user who opens the drawer and taps Home will round-trip back to `/sign-in` — harmless but worth hiding for unauth like the existing `/shifts` filter does. CI: Playwright job still pending at time of review.

PR #340, opened 2026-05-25
feat(shifts): show CSP on /shifts list + /shifts/[timeId]/volunteers

### LINKED ISSUE #339: Show CSP as a column on /shifts list page
From Mew (Discord, 2026-05-25, after debugging the "21 CSP feels high" report):

> we need to provide people with how many CSP each position on each shift is worth. ... this should appear as another column on the shifts page

## Current state
`/shifts` (`Shifts.tsx`) lists shifts in a table. Columns today are roughly:
- date, dateName, type, startTime/endTime, slotsFilled / slotsTotal

The volunteer doesn't see CSP until they're already signed up and visit `/info`. The position picker on `/shifts/[timeId]/volunteers` also doesn't show CSP yet — but Mew's request is the **list page**, not the picker.

## Why
Volunteers need to plan toward a CSP target (12 for the base SAP, more for some roles). Same-duration shifts can have very different CSP values (e.g. 2-hour Tech Support = 4 CSP vs. 2-hour 15:00-17:00 OpenSun = 2 CSP), so they're picking blind without it.

## Wrinkle
CSP is **per position** within a shift, not per shift overall. Each shift can have multiple position rows with different `sap_points`. Options for the column:

1. **Range**: `4 CSP` if all positions agree, `2-4 CSP` if they differ.
2. **Min/Max**: `2-4` (compact, requires legend).
3. **Sum**: total points if you took all slots — meaningless to most volunteers, skip.
4. **Average**: misleading.

(1) reads best. Needs API to return either the unique sap_points set per shift, or the min/max pair.

## Implementation sketch
- `client/src/utils/getShiftList.ts` already aggregates per-shift; extend to track `sapPointsMin` / `sapPointsMax` from the SQL rows.
- `client/src/pages/api/shifts/index.ts` SQL needs `stp.sap_points` selected (currently aggregates slots but not points — check).
- `client/src/components/types/shifts.ts` `IResShiftRowItem` adds the new fields.
- `Shifts.tsx` adds the column.

## Out of scope (separate issue if wanted)
- Showing per-position CSP on the position picker dialog when signing up — was my first read of Mew's request; superseded by this clarification.

## Acceptance
- `/shifts` shows a "CSP" column alongside slot counts.
- Format: single number if all positions agree, range like `2-4` if they don't.
- No regressions in existing slot-count math (Map-keyed dedupe from #318 stays correct).

mbhewitt:
Closes #339.

## Summary

Volunteers couldn't see CSP value until *after* sign-up — they had to take a position blind, then check `/info` to learn what it was worth. Mew flagged this after debugging a \"21 CSP feels high\" report; turned out the math was right but the visibility was missing.

Two surfaces get the new column:

1. **`/shifts`** (list page) — new \"CSP\" column. Renders a single number when every position on the shift agrees (the common case), or a range like `2-4` when they differ. Format chosen by Mew.

2. **`/shifts/[timeId]/volunteers`** — the positions table at the top gets a per-position \"CSP\" column alongside the existing Name and Filled / Total columns.

## Files
- `client/src/pages/api/shifts/index.ts` — both training and non-training SELECTs now fetch `stp.sap_points`.
- `client/src/pages/api/shifts/[timeId]/volunteers/index.ts` — position SELECT fetches `stp.sap_points`; payload exposes it as `csp`.
- `client/src/utils/getShiftList.ts` — track `cspMin` / `cspMax` per shift while iterating the LEFT-JOIN row stream. Sentinels normalized to 0 in the (defensive) no-position case.
- `client/src/components/types/shifts/index.ts` — additive: `IResShiftRowItem` gains `cspMin`/`cspMax`; `IResShiftPositionCountItem` gains `csp`. Existing destructuring callsites compile unchanged.
- `client/src/app/shifts/Shifts.tsx` — column def + cell formatter (single vs range).
- `client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx` — column def + cell formatter.

## Out of scope
- The position-picker dropdown (`ShiftVolunteersDialogAdd.tsx`) doesn't yet show CSP — Mew's explicit ask was the list page and the positions table, not the picker. If wanted, that's a small follow-up using the same `csp` field already exposed by the API.

## Test plan
- [ ] `/shifts` shows a \"CSP\" column. Shifts where all positions share `sap_points` show one number (e.g. `4`); shifts with varying values show a range (e.g. `2-4`).
- [ ] `/shifts/<timeId>/volunteers` positions table shows a \"CSP\" column matching each position's `sap_points`.
- [ ] Existing slot-count math unchanged (Map-keyed dedupe from #318 stays correct — verified by reading the modified getShiftList against the existing tests' expected shape).

### DIFF
--- a/client/src/app/shifts/Shifts.tsx
+++ b/client/src/app/shifts/Shifts.tsx
@@ -125,6 +125,13 @@ export const Shifts = () => {
         sort: false,
       },
     },
+    {
+      name: "CSP",
+      options: {
+        filter: false,
+        sort: false,
+      },
+    },
   ]);
 
   // fetching, mutation, and revalidation
@@ -242,6 +249,8 @@ export const Shifts = () => {
   const colorMapDisplay = getColorMap(data);
   const dataTable = data.map(
     ({
+      cspMax,
+      cspMin,
       date,
       dateName,
       department: { name: departmentName },
@@ -252,6 +261,8 @@ export const Shifts = () => {
       startTime,
       type,
     }) => {
+      const cspDisplay =
+        cspMin === cspMax ? `${cspMin}` : `${cspMin}-${cspMax}`;
       return [
         id, // hide for row click
         date, // hide for filter dialog (Present/Future vs Past)
@@ -264,6 +275,7 @@ export const Shifts = () => {
           sx={{ backgroundColor: colorMapDisplay[departmentName] }}
         />,
         `${slotsFilled} / ${slotsTotal}`,
+        cspDisplay,
       ];
     }
   );
--- a/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
+++ b/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
@@ -378,10 +378,22 @@ export const ShiftVolunteers = ({
         sort: false,
       },
     },
+    {
+      name: "CSP",
+      options: {
+        filter: false,
+        sort: false,
+      },
+    },
   ];
   const dataTablePositions = dataShiftVolunteersItem.positionList.map(
-    ({ positionName, slotsFilled, slotsTotal }: IResShiftPositionCountItem) => {
-      return [positionName, `${slotsFilled} / ${slotsTotal}`];
+    ({
+      csp,
+      positionName,
+      slotsFilled,
+      slotsTotal,
+    }: IResShiftPositionCountItem) => {
+      return [positionName, `${slotsFilled} / ${slotsTotal}`, csp];
     }
   );
   const optionListCustomPositions = {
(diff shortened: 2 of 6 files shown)


mbhewitt:
LGTM. Aggregation in getShiftList.ts correctly sits inside the timePositionSeen guard so per-position rows multiplied by the LEFT JOIN with op_volunteer_shifts only update cspMin/cspMax once. The Infinity/-Infinity sentinel normalization is a sound safety net but unreachable in practice since both /api/shifts queries INNER JOIN op_shift_time_position, guaranteeing at least one position row per shift_times_id. Type changes are additive; only the API handler constructs IResShiftPositionCountItem and IResShiftRowItem literals (both updated to populate the new fields), all consumer sites destructure subsets. Shifts.tsx correctly destructures cspMin/cspMax and renders the range vs single-value display. CI Playwright still pending at time of review.

PR #341, opened 2026-05-25
ui(shifts): move CSP column before Type + pink/bold filled count if open

mbhewitt:
Two small tweaks from Mew (2026-05-25, follow-up to #340):

## Changes

1. **CSP column position**: was rightmost, now sits between Time and Type. Column index reordering only touches positions 4-7 in the dataTable row; indexes 0 (id for row click), 1 (hidden date for sort), and 2 (display date for row alternation) are untouched, so existing `sortCompare` / `setRowProps` / `onRowClick` keep working.

2. **Filled count emphasis**: when `filled < total`, the filled portion of the cell renders in **pink + bold** (e.g. `**2** / 5`). Full shifts (`filled >= total`) render plain. Uses `customBodyRender` so the underlying string `"<filled> / <total>"` is preserved for the Full / Open filter logic that splits on `" / "`.

## Pink reference
`theme.palette.secondary.main` — Census brand pink (#EA008B per the Standards doc).

## Test plan
- [ ] /shifts column order: Date / Time / **CSP** / Type / Filled / Total.
- [ ] Open shift (e.g. 0 / 5): "0" is pink and bold; " / 5" is plain.
- [ ] Full shift (e.g. 5 / 5): plain text.
- [ ] Full / Open filter still works.
- [ ] Row click still navigates to /shifts/<id>/volunteers (column index 0 untouched).

### DIFF
--- a/client/src/app/shifts/Shifts.tsx
+++ b/client/src/app/shifts/Shifts.tsx
@@ -79,6 +79,13 @@ export const Shifts = () => {
       options: { filter: false, sortThirdClickReset: true },
     },
     {
+      name: "CSP",
+      options: {
+        filter: false,
+        sort: false,
+      },
+    },
+    {
       name: columnNameTypeHidden, // hide for filter dialog
       label: "Type",
       options: {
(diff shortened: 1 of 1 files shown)


mbhewitt:
LGTM.

Verified index-dependent logic against new column order (id=0, dateHidden=1, Date=2, Time=3, CSP=4, TypeHidden=5, Type=6, Filled/Total=7):
- sortCompare reads rowData[1] = hidden date string, correct.
- setRowProps reads row[2] = formatted Date display, correct.
- onRowClick reads row[0] = id, correct.

customBodyRender returns the raw `value` string when filled >= total and a JSX span otherwise; mui-datatables filter logic operates on the underlying data array, not the rendered output, so the " / " split in filterOptions.logic stays intact.

theme is captured by closure in the useImmer initializer (line 37) before useTheme() is called (line 143), but the closure is only invoked downstream during DataTable render after theme is initialized, so no TDZ at call time. theme.palette.secondary.main resolves to COLOR_CENSUS_PINK per layout.tsx.

CI: Playwright pending; no blockers from review.

PR #349, opened 2026-05-25
ui(info): Staff exempt from CSP — rewrite copy + auto-complete role thresholds

mbhewitt:
Per Mew (Discord, 2026-05-25):

> On the check list for staff it says, "As Census staff, your early entry is handled separately. Work your butt off and have a great burn!" It should say "As Census Staff we know you will work your butt off, so we dont track CSP." And anywhere you see CSP in the checklist such as for camp, just put that line and regardless if they meet the csp mark it complete for any that require a CSP check.

## Two changes
1. **Staff bypass item** — copy updated to the new "we don't track CSP" line.
2. **Role-threshold loop** (Counter Culture / Census Lab / Census Ticket) — short-circuit for Staff: emit a "Requirements met" item with the no-CSP copy and `done: true`, regardless of actual `currentCsp`. Non-staff path unchanged.

Shared copy hoisted to a local `staffNoCspCopy` Typography so the two sites stay in sync.

(SAP Requirements block above already skips Staff via the existing `!isStaff` guard — no change needed there.)

## Test plan
- [ ] Staff volunteer view `/info`: "Staff — Early entry confirmed" item shows new copy.
- [ ] Staff volunteer with a camp/ticket role (e.g. Counter Culture, Census Lab, Census Ticket) sees "Requirements met" for that item with the same no-CSP copy.
- [ ] Non-staff volunteer with the same camp/ticket role still sees CSP progress bar and old behavior.

### DIFF
--- a/client/src/app/volunteers/[shiftboardId]/info/VolunteerInfo.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/info/VolunteerInfo.tsx
@@ -723,8 +723,12 @@ export const VolunteerInfo = ({ shiftboardId }: IVolunteerInfoProps) => {
   const completedItems = checklistItems.filter((item) => item.done);
 
   // show CSP bar when relevant
+  // Staff are CSP-exempt across the whole checklist (per Mew 2026-05-25),
+  // so the global gauge should not appear at all for them — even if they
+  // happen to have a role threshold attached.
   const showCspBar =
-    (isPreOpen && !hasOtherSap && !isStaff) || roleThresholds.length > 0;
+    !isStaff &&
+    ((isPreOpen && !hasOtherSap) || roleThresholds.length > 0);
 
   // render
   // ------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


mbhewitt:
LGTM on the checklist changes. Verified: `staffNoCspCopy` is hoisted above the staff-bypass block and reused in both the staff item and the role short-circuit; the short-circuit is gated by `if (isStaff)` with `continue` so non-staff falls through to the existing CSP-bar branch; `displayRole` was correctly hoisted above the gate so both branches can use it. One non-blocker: `showCspBar` on line 711 is still `(isPreOpen && !hasOtherSap && !isStaff) || roleThresholds.length > 0`, so a Staff volunteer who holds Counter Culture / Lab / Ticket will still see the global CSP bar render even though their checklist items now say no-CSP. If the intent was "never sees a CSP gauge anywhere," add `&& !isStaff` to the second clause too. Playwright pending.

mbhewitt:
Reviewer caught: the global `showCspBar` on line 727 could still evaluate true for staff with a role threshold. Pushed a follow-up commit (62719c3) to hoist `!isStaff` to the top of the condition so staff never see CSP UI anywhere on /info.

Issue #350, opened 2026-05-25
Tighten passcode PATCH auth: any signed-in user can change anyone's passcode


mbhewitt:
Surfaced while building the reveal-passcode flow (2026-05-25).

## Bug
`client/src/pages/api/volunteers/[shiftboardId]/account/passcode/index.ts` PATCH handler is wrapped in `withAuth`, which proves the caller has a valid session but does NOT verify that the caller is either:
- the volunteer themselves (matching `session.shiftboardId === req.query.shiftboardId`), or
- an admin/super-admin.

So any signed-in user can today fire `PATCH /api/volunteers/<any-id>/account/passcode` and overwrite that volunteer's passcode.

## Fix
Inside the PATCH branch, before the UPDATE:
```ts
const requestId = Number(req.query.shiftboardId);
const isSelf = session.shiftboardId === requestId;
const isAdmin = /* read session roles, check ROLE_ADMIN_ID or ROLE_SUPER_ADMIN_ID */;
if (!isSelf && !isAdmin) {
  return res.status(403).json({ statusCode: 403, message: "Forbidden" });
}
```

`withAuth` may need to be extended to expose the session's role list (or we re-query op_volunteer_roles inside the handler).

## Out of scope
- The new self-only GET endpoint (separate PR for the reveal-passcode flow) already does the right check from the start, since it's net-new code.
- Broader audit of other `withAuth`-wrapped endpoints with the same gap. Likely candidates: anything under `/api/volunteers/[shiftboardId]/...` that mutates state.

## Severity
Medium. Requires a valid Census session to exploit (signed-in user, not anonymous), but bypasses any expected per-volunteer privacy boundary. Should be fixed before any external pen-test.

PR #351, opened 2026-05-25
feat(passcode): auto-generate for new Okta accounts + self-only reveal endpoint

mbhewitt:
Backend half of the on-playa passcode UX work (per Mew 2026-05-25). UI follow-up in a separate PR — drops "Reveal Passcode" button on the /info Security section + Census-logo image with digits embedded.

Filed adjacent: #350 (PATCH auth gap on the same endpoint — not fixed here per Mew's explicit "admin can still set someone's passcode" scope).

## Three pieces

1. **New helper** `client/src/utils/generatePasscode.ts` — 4-digit passcode via `crypto.randomInt`, leading zeros preserved (e.g. `"0042"`).

2. **Okta callback** (`callback.ts:312-327`) — when creating a brand-new volunteer row (no prior shiftboard_id, no existing email match), also INSERT a random passcode so the user can sign in on-playa without an admin having to set one for them.

3. **New GET method** on `/api/volunteers/[shiftboardId]/account/passcode` — returns `{ passcode }` for self-callers only. **403 for anyone else, including admins.** The existing PATCH method is untouched.

## Backfill (post-deploy)

4 existing volunteers on prod have no passcode. After deploy:
```sql
UPDATE op_volunteers
SET passcode = LPAD(FLOOR(RAND()*10000), 4, '0'),
    update_volunteer = true
WHERE passcode IS NULL OR passcode = '';
```
Run as a one-off SQL, not as part of this PR.

## Test plan
- [ ] New Okta signup → new `op_volunteers` row has a 4-digit `passcode`.
- [ ] GET `/api/volunteers/<own-id>/account/passcode` as self → 200 with `{ passcode }`.
- [ ] GET `/api/volunteers/<other-id>/account/passcode` as different volunteer → 403.
- [ ] GET as admin viewing someone else → 403 (admin reveal not allowed per spec).
- [ ] Existing PATCH `/api/volunteers/<id>/account/passcode` unchanged in behavior.

### DIFF
--- a/client/src/pages/api/auth/okta/callback.ts
+++ b/client/src/pages/api/auth/okta/callback.ts
@@ -3,6 +3,7 @@ import type { NextApiRequest, NextApiResponse } from "next";
 
 import { buildSessionCookie } from "@/lib/session";
 import { generateId } from "@/utils/generateId";
+import { generatePasscode } from "@/utils/generatePasscode";
 import { pool } from "lib/database";
 
 interface OktaTokenResponse {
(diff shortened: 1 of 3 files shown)


mbhewitt:
LGTM. `generatePasscode` correctly pads `randomInt(0, 10_000)` to a 4-char string (e.g. 7 -> "0007"). GET handler pulls `shiftboardId` from the `withAuth`-injected `session` rather than the cookie/query, and `Number(req.query.shiftboardId) !== session.shiftboardId` correctly bridges the URL string vs numeric session id. PATCH callers (`PasscodeDialogUpdate.tsx`) are isolated by the `req.method` switch. `import { randomInt } from "crypto"` is fine in the Node API runtime (equivalent to `node:crypto`). The `!session` guard inside the GET case is dead code since `withAuth` already 401s, but harmless. Non-blocking nit: PATCH is still ungated here — tracked in #350 per the comment, so out of scope. Playwright check still pending at time of review.

PR #353, opened 2026-05-25
ui(shifts): suppress empty Details/Meal/Notes rows on shift page

### LINKED ISSUE #234: Suppress shift detail row titles (left column) if no details (right column) are available to display for a given shift
Shift details say "Meal" for shifts where no meal is provided, as the table at the top of the shift info page includes a row titled "Meal" in the left column. This row is currently displayed even for shifts which do not have meal detail available to include in the detail column on the right. This could cause onfusion for volunteers who view the inclusion of the word "Meal" as indication that a meal will be awarded for this shift.

Image

mbhewitt:
Closes #234.

Per Chipper (Discord, 2026-05-25):

> on the shift volunteers page there's always a table at the top that shows The shift name and then below it it says meal and then below that is notes. The whole table still shows even if there's nothing to show for meal or notes. I think we should suppress any rows Where the second column is not filled. Especially for the meals, let's not rub it in that they don't get a meal for that particular shift.

## Change
`client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx` — replace the static three-row Details/Meal/Notes JSX with a filtered list driven by the shift payload. Dividers only render between populated rows. If all three are empty the whole Card collapses (no orphan header).

Adds `Fragment` to the existing `react` import for the row+divider key wrapper.

## Test plan
- [ ] Shift with all three fields populated: looks identical to today.
- [ ] Shift with Details only (no meal, no notes): renders just the Details row.
- [ ] Shift with no Details / Meal / Notes: Card collapses entirely, no orphan card.
- [ ] Whitespace-only meal (just spaces): treated as empty, not shown.

### DIFF
--- a/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
+++ b/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
@@ -29,7 +29,7 @@ import {
 import dayjs from "dayjs";
 import Link from "next/link";
 import { useSnackbar } from "notistack";
-import { useContext, useEffect, useState } from "react";
+import { Fragment, useContext, useEffect, useState } from "react";
 import { io } from "socket.io-client";
 import useSWR, { KeyedMutator } from "swr";
 import useSWRMutation from "swr/mutation";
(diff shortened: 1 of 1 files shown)


mbhewitt:
LGTM. Fragment import added, filter correctly handles null/undefined/whitespace (these are string fields so the truthy guard losing literal 0/false is moot), divider only between rows, Card returns null cleanly when no rows have content. The inline IIFE is fine for a three-row table; if this grows I'd lift detailRows to a const above the return, but not a blocker. CI still pending Playwright — happy to merge once green.

PR #358, opened 2026-05-25
feat(passcode-reveal): self-only Reveal button + Census-logo PNG embed

mbhewitt:
UI half of the passcode UX work (per Mew 2026-05-25). Backend shipped in #351 + the backfill SQL ran today.

## Component
New `PasscodeReveal` in `client/src/app/volunteers/[shiftboardId]/account/`. Lives in the Security section on `/info`, **only when the viewer is looking at their own account** (`session.shiftboardId === url shiftboardId`). Admins viewing someone else's `/info` don't get the button — they can still reset via the existing Update Passcode dialog right next to it, just can't read the existing value.

## Flow
1. Volunteer clicks "Reveal passcode".
2. `fetch GET /api/volunteers/<id>/account/passcode` (the self-only endpoint added in #351).
3. Canvas loads `/general/logo-census.png` and overlays a pink rectangle on top of the "CENSUS" wordmark, then stamps the 4 digits in chunky sans-serif in the same vertical band.
4. Renders at 260×260 CSS pixels (canvas backing is the logo's native 1019×1015 so the digits stay crisp).
5. Helper caption: *"This passcode is for signing in on tablets on-playa only."*
6. "Hide" button collapses back to the reveal-button state.

## Notes
- Mask coordinates are tuned to the existing PNG. If the logo asset changes those need re-tuning. Could switch to SVG later for resolution independence — not needed right now.
- The reveal endpoint is self-only at the API layer too (#351), so even with the button hidden in the UI, admins still can't fetch.

## Test plan
- [ ] Sign in as a volunteer, visit `/volunteers/<own-id>/info`, scroll to Security. See "Reveal passcode" button next to "Update passcode".
- [ ] Click → canvas renders the Census logo with the 4-digit passcode where "CENSUS" used to be.
- [ ] "Hide" button collapses back.
- [ ] Sign in as an admin, visit `/volunteers/<other-id>/info`. No Reveal button. "Update passcode" button still there.
- [ ] Fetch `GET /api/volunteers/<other-id>/account/passcode` directly as admin → 403 (verifies API layer matches UI).

### DIFF
--- a/client/src/app/volunteers/[shiftboardId]/account/PasscodeReveal.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/account/PasscodeReveal.tsx
@@ -1,0 +1,132 @@
+"use client";
+
+import {
+  Visibility as VisibilityIcon,
+  VisibilityOff as VisibilityOffIcon,
+} from "@mui/icons-material";
+import { Button, Stack, Typography } from "@mui/material";
+import { useEffect, useRef, useState } from "react";
+
+interface IPasscodeRevealProps {
+  shiftboardId: number;
+}
+
+// Render the volunteer's passcode embedded in the Census logo. Per Mew
+// 2026-05-25: replace the "CENSUS" wordmark with the 4-digit passcode so
+// it reads as a delight artifact rather than a plain text reveal.
+//
+// Coordinates below are tuned to logo-census.png (1019×1015 native). The
+// pink wash matches the brand pink (#EA008B) so the rectangle blends
+// into the badge background.
+const LOGO_SRC = "/general/logo-census.png";
+const PINK = "#EA008B";
+const MASK_X = 195;
+const MASK_Y = 460;
+const MASK_W = 630;
+const MASK_H = 130;
+
+export const PasscodeReveal = ({ shiftboardId }: IPasscodeRevealProps) => {
+  const [passcode, setPasscode] = useState<null | string>(null);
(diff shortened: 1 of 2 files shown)


mbhewitt:
LGTM with minor non-blocking notes. Verified: server-side self-only gate is in place in `pages/api/volunteers/[shiftboardId]/account/passcode/index.ts` (GET returns 403 unless `session.shiftboardId === Number(req.query.shiftboardId)`), so admin direct-fetch is blocked even if the UI were wrong. Type-equality on `isSelfView` is sound — `VolunteerInfo` is called with `shiftboardId={Number(shiftboardId)}` from `page.tsx`, and `sessionInitial.user.shiftboardId` is `0` (number), so the strict `===` is number-vs-number. Two small flags for follow-up, not blockers: (1) the canvas `useEffect` has no cleanup — under StrictMode double-mount, the first effect's `img.onload` can fire after unmount and write to a now-detached canvas (harmless second write wins, but a tiny stale-write/leak; an `isCancelled` flag or `img.onload = null` in a cleanup would be cleaner); (2) doc drift — the comment says logo is 1019×1015 but the file is actually 1008×1008; mask coords still land roughly over the wordmark but the pink fill will leave a visible seam against the black interior of the C-cutout, so the digits may read as a pink block on black rather than blending — worth an eyeball on the deploy before iterating. CI: Playwright still pending at review time.

PR #359, opened 2026-05-25
ui(passcode-reveal): move on-playa-only copy to header + iterate logo mask

mbhewitt:
Iteration on #358 per Mew (Discord, 2026-05-25):

> move the passcode description to the main part where it currently says just "Passcode" ... 2 leave the C in Census and move the numbers over and make them a touch larger.

## Changes
1. **Helper copy** moves from the canvas-revealed state into a `body2` sub-label under "Passcode" in the Security card header. Now visible whether or not the user has clicked Reveal.
2. **Mask start** shifts right (MASK_X 195 → 305) so the "C" of CENSUS stays visible. Mask width drops accordingly (630 → 520). Badge reads as "C ####".
3. **Digit font** bumps 160 → 180 to fill the remaining space better.

No other behavior changes — still self-only, still client-side canvas, still Hide button to collapse.

## Test plan
- [ ] /info Security section: "This passcode is for signing in on tablets on-playa only." is always visible under "Passcode" (no reveal needed).
- [ ] Click Reveal → canvas renders the badge with "C ####" visible instead of "####".
- [ ] Digits look comfortably sized inside the masked area (not crowded, not floating).
- [ ] Hide button still collapses back.

### DIFF
--- a/client/src/app/volunteers/[shiftboardId]/account/PasscodeReveal.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/account/PasscodeReveal.tsx
@@ -15,15 +15,17 @@ interface IPasscodeRevealProps {
 // 2026-05-25: replace the "CENSUS" wordmark with the 4-digit passcode so
 // it reads as a delight artifact rather than a plain text reveal.
 //
-// Coordinates below are tuned to logo-census.png (1019×1015 native). The
-// pink wash matches the brand pink (#EA008B) so the rectangle blends
-// into the badge background.
+// Coordinates below are tuned to logo-census.png (1019×1015 native).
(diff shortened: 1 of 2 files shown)


mbhewitt:
LGTM. Typography import still used by the error caption (line 102). Helper text is unconditional — lives in the Passcode label column, not behind isSelfView/isAdmin. MASK_X=305 / MASK_W=520 cleanly preserves the 'C' (110px = one letter slot from the original 630px wordmark) and the 4 digits get 130px each in the 130px-tall band, so 180px Arial Black fits the intended chunky look. Playwright still pending — assume green and ship if it passes.

PR #384, opened 2026-06-01
feat(mail): email unsubscribe — role-based flag + footer + info-page checkbox

mbhewitt:
## Summary

Adds the email unsubscribe loop now that the queue is sending automated mail.

- **Storage**: an `EmailUnsubscribed` role on `op_volunteer_roles` (role_id 2000020, seeded in [migrations/003](migrations/003_email_unsubscribed_role.sql)). Lives next to every other per-volunteer flag (`OtherSAP`, `BurnerProfileUpdated`, `BehavioralStandards`). No schema change.
- **Enforcement**: `enqueueEmail` gets an optional `recipientShiftboardId`. When set, [queue.ts](client/lib/mail/queue.ts) checks the role and skip-returns `{ id: 0, skipped: true }`; otherwise it appends a text + HTML footer linking to `${APP_BASE_URL}/unsubscribe`. Threaded through `notifyAssignment` / `notifyRemoval` / `notifyRestoration`. Critical-drop (VC list) and contact-form sends intentionally don't pass it — opt-out doesn't apply to those.
- **One-click unsub**: [`/unsubscribe`](client/src/app/unsubscribe/page.tsx) is wrapped in `AuthGate` so unauthenticated clicks land on `/sign-in?returnTo=/unsubscribe` (Okta) and come back. Auto-fires the POST exactly like `/training/confirmation/[code]`, then `router.replace` to `/volunteers/[id]/info#settings`.
- **Checkbox**: `GET /api/volunteers/[id]/info` returns `emailUnsubscribed`; new `POST .../info/email-preference` (self-only, mirrors `other-sap`) toggles the role. `VolunteerInfo.tsx` adds a `<Box id="settings">` Settings section at the bottom with a "Receive emails from Census" checkbox — `#settings` is the redirect anchor.

## Test plan

- [ ] Apply `migrations/003_email_unsubscribed_role.sql` on test DB; confirm role_id=2000020 row in `op_roles`
- [ ] Sign in, visit `/volunteers/<self-id>/info`, scroll to Settings, uncheck the box → snackbar "Email preference updated"
- [ ] Trigger an assignment email (add a shift) → footer present, link goes to `/unsubscribe`
- [ ] Trigger another assignment to a still-subscribed account → footer present
- [ ] Re-check the Settings box → next assignment notify enqueues normally
- [ ] Uncheck again, then trigger an assignment → check `op_email_queue` shows no new row (returned `skipped: true`)
- [ ] Click the unsubscribe link while signed out → Okta sign-in, then auto-unsub, lands on `/volunteers/<id>/info#settings` with checkbox unchecked
- [ ] Click the unsubscribe link while signed in → no sign-in detour
- [ ] Verify critical-drop email (drop a critical shift) → no footer, sends to VC list normally
- [ ] Verify contact form email → no footer

### DIFF
--- a/client/lib/mail/index.ts
+++ b/client/lib/mail/index.ts
@@ -39,7 +39,9 @@ function config(): MailConfig {
   return runtimeConfig;
 }
 
-export async function enqueueEmail(args: EnqueueArgs): Promise<{ id: number }> {
+export async function enqueueEmail(
+  args: EnqueueArgs
+): Promise<{ id: number; skipped?: boolean }> {
   return enqueueEmailRaw(pool, config(), args);
 }
 
--- a/client/lib/mail/queue.ts
+++ b/client/lib/mail/queue.ts
@@ -4,19 +4,73 @@ import type { ResultSetHeader, RowDataPacket } from "mysql2";
 import { nextAttemptDelaySeconds } from "./backoff";
 import type { EmailRow, EnqueueArgs, MailConfig } from "./types";
 
+// Kept local to the mail layer so queue.ts has no dep on src/constants.ts
+// (which is React-side). Must stay in sync with
+// ROLE_EMAIL_UNSUBSCRIBED_ID in client/src/constants.ts and the seed in
+// migrations/003_email_unsubscribed_role.sql.
+const ROLE_EMAIL_UNSUBSCRIBED_ID = 2000020;
+
+const APP_BASE_URL =
+  process.env.APP_BASE_URL ?? "https://volunteers.census.burningman.org";
+
 function joinAddrs(v: string | string[] | undefined): string | null {
   if (v === undefined) return null;
   if (Array.isArray(v)) return v.join(", ");
   return v;
 }
 
+async function isUnsubscribed(
+  pool: Pool,
+  shiftboardId: number
+): Promise<boolean> {
+  const [rows] = await pool.query<RowDataPacket[]>(
+    `SELECT 1
+       FROM op_volunteer_roles
+      WHERE shiftboard_id = ?
+        AND role_id = ?
+        AND add_role = true
+        AND remove_role = false
+      LIMIT 1`,
+    [shiftboardId, ROLE_EMAIL_UNSUBSCRIBED_ID]
+  );
+  return rows.length > 0;
+}
+
+function appendUnsubscribeFooter(
+  bodyText: string,
+  bodyHtml: string | undefined,
+  unsubscribeUrl: string
+): { bodyText: string; bodyHtml: string | undefined } {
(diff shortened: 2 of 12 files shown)


mbhewitt:
**Independent review — reads cold from the diff.**

Overall this is clean and follows the existing patterns (other-sap, profile-updated). No blockers found. A few notes:

**Correctness — checked, OK:**
- `queue.ts:27-37` (`add_role=true AND remove_role=false`) and the toggle endpoint (`email-preference/index.ts:36`) agree: ON writes `[true,false]`, OFF writes `[false,true]`. Re-subscribe and re-unsubscribe both UPDATE the existing row, not insert dups (PK is `(shiftboard_id, role_id)`).
- 2000020 is unused anywhere else in the repo (`grep -rn 2000020` only finds the four new sites + the seed). The 2000xxx status-flag range convention matches the comment in `migrations/003`.
- Footer + `MAIL_OVERRIDE_TO` banner can't double up: footer is appended at enqueue, banner is prepended at send (`transport.ts:43-49`). Independent stages, no overlap.
- `Number(shiftboardId) !== session.shiftboardId` self-check (`email-preference/index.ts:24`) is safe against array/undefined: `Number([...])` → NaN, `NaN !== n` → true → 403. Good.
- `IResVolunteerInfo.emailUnsubscribed` — only destructured at `VolunteerInfo.tsx:451`. Nothing uses `Required<>` or exhaustive checks, so adding the field is backward compatible.
- No existing mail-unit tests touch `enqueueEmail`, so the return-shape change (`{id} → {id, skipped?}`) doesn't break anything (`mail-worker.test.ts` only exercises `tickOnce`).

**Minor / non-blocking:**
- Two copies of `const ROLE_EMAIL_UNSUBSCRIBED_ID = 2000020` (`queue.ts:11`, `info/index.ts:39`, `email-preference/index.ts:12`) plus the `src/constants.ts:94` export. The `queue.ts` duplication is justified in-comment (no React-side import), but `info/index.ts` and `email-preference/index.ts` could pull from `@/constants`. Same drift risk the other `ROLE_*_ID` constants already have — fine to leave.
- `info/index.ts:115-122` reads roles with `WHERE remove_role=false` (so `[false,false]` counts as having the role), while the queue uses `add_role=true AND remove_role=false`. In practice the toggle endpoint never writes `[false,false]`, so the UI and the mailer agree — but if some other process ever seeded such a row, the checkbox would show \"unsubscribed\" while mail still flowed. Tightening the GET to `add_role=true AND remove_role=false` would close that gap.
- `migrations/003`: `ON DUPLICATE KEY UPDATE` on `op_roles` matches PK `role_id` first. If 2000020 ever held a different role name, this silently renames it. Worth flagging in case of staging weirdness, but unlikely in practice.
- `Unsubscribe.tsx:50` guards on `!shiftboardId`, which also early-returns on `0`. Not a real risk (no volunteer has id 0) but worth noting.

Looks good — recommend merging.

PR #388, opened 2026-06-01
fix(utils): make generateId properly async

mbhewitt:
## Summary
- `generateId()` was declared sync but ran its DB duplicate-check inside an inner async helper that was never awaited. It always returned `0`. Every new-volunteer INSERT in the Okta callback was writing `shiftboard_id=0`; two concurrent first-time logins would collide on the PK and one would 500. The other 8 callers (admin create routes for roles, dates, shift categories, positions, shift types) had the same silent zero behavior — first INSERT succeeds with id 0, every subsequent attempt fails the duplicate check.
- Rewrites `generateId` as a properly async function with a `while` loop in place of the unbounded recursion, and adds `await` at all 9 call sites.

## Why a loop instead of awaited recursion
The original recursion was unbounded and ate stack on collision; a loop is the same semantics with no recursion. Functionally identical, easier to reason about.

## Scope notes
- I noticed `client/src/pages/api/shifts/types/index.ts` has unrelated fire-and-forget bugs (`timeList.forEach(async ...)` and a bare `handleTimeListAdd(...)` call without await). Out of scope for this PR — opening separately if you want.
- Pre-existing: `npm run lint` fails because Next 16 removed `next lint`. `npx tsc --noEmit` and `npm run build` both pass cleanly with this change.

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npm run build` — clean
- [ ] Manual: sign in via Okta as a brand-new email that's NOT in `sb_pinfo` and NOT in `op_volunteers`; confirm the row written has a non-zero `shiftboard_id`
- [ ] Manual: hit `POST /api/roles` (or any admin create route) and confirm the new row has a non-zero id

### DIFF
--- a/client/src/pages/api/auth/okta/callback.ts
+++ b/client/src/pages/api/auth/okta/callback.ts
@@ -372,7 +372,7 @@ const oktaCallback = async (req: NextApiRequest, res: NextApiResponse) => {
 
     // 4. create new volunteer with a generated id
     if (!shiftboardId) {
-      const newId = generateId(`
+      const newId = await generateId(`
         SELECT shiftboard_id
         FROM op_volunteers
         WHERE shiftboard_id=?
--- a/client/src/pages/api/dates/index.ts
+++ b/client/src/pages/api/dates/index.ts
@@ -38,7 +38,7 @@ const calendar = async (req: NextApiRequest, res: NextApiResponse) => {
     case "POST": {
       // create date
       const { date, name } = JSON.parse(req.body);
-      const dateIdNew = generateId(
+      const dateIdNew = await generateId(
         `SELECT date_id
         FROM op_dates
         WHERE date_id=?`
--- a/client/src/pages/api/roles/index.ts
+++ b/client/src/pages/api/roles/index.ts
@@ -38,7 +38,7 @@ const roles = async (req: NextApiRequest, res: NextApiResponse) => {
     case "POST": {
       // create role
       const { name } = JSON.parse(req.body);
-      const roleIdNew = generateId(
+      const roleIdNew = await generateId(
         `SELECT role_id
         FROM op_roles
         WHERE role_id=?`
--- a/client/src/pages/api/shifts/categories/index.ts
+++ b/client/src/pages/api/shifts/categories/index.ts
@@ -46,7 +46,7 @@ const shiftCategories = async (req: NextApiRequest, res: NextApiResponse) => {
         department: { name: departmentName },
         name: categoryName,
       }: IReqShiftCategoryItem = JSON.parse(req.body);
-      const shiftCategoryIdNew = generateId(
+      const shiftCategoryIdNew = await generateId(
         `SELECT shift_category_id
         FROM op_shift_category
         WHERE shift_category_id=?`
--- a/client/src/pages/api/shifts/positions/index.ts
+++ b/client/src/pages/api/shifts/positions/index.ts
@@ -50,7 +50,7 @@ const shiftPositions = async (req: NextApiRequest, res: NextApiResponse) => {
         role: { id: roleId },
         startTimeOffset,
       }: IReqShiftPositionItem = JSON.parse(req.body);
-      const positionIdNew = generateId(
+      const positionIdNew = await generateId(
         `SELECT position_type_id
         FROM op_position_type
         WHERE position_type_id=?`
--- a/client/src/pages/api/shifts/types/[typeId]/index.ts
+++ b/client/src/pages/api/shifts/types/[typeId]/index.ts
@@ -508,7 +508,7 @@ const shiftTypeUpdate = async (req: NextApiRequest, res: NextApiResponse) => {
             );
             // else insert them into the table
           } else {
-            const timePositionIdNew = generateId(
+            const timePositionIdNew = await generateId(
               `SELECT time_position_id
               FROM op_shift_time_position`
             );
--- a/client/src/pages/api/shifts/types/index.ts
+++ b/client/src/pages/api/shifts/types/index.ts
@@ -21,7 +21,7 @@ export const handleTimeListAdd = async ({
   // insert new shift time rows
   timeList.forEach(
     async ({ endTime, instance, meal, notes, positionList, startTime }) => {
-      const timeIdNew = generateId(
+      const timeIdNew = await generateId(
         `SELECT shift_times_id
         FROM op_shift_times`
       );
@@ -56,7 +56,7 @@ export const handleTimeListAdd = async ({
       );
 
       positionList.forEach(async ({ alias, positionId, sapPoints, slots }) => {
-        const timePositionIdNew = generateId(
+        const timePositionIdNew = await generateId(
           `SELECT time_position_id
           FROM op_shift_time_position`
         );
@@ -125,7 +125,7 @@ const shiftTypes = async (req: NextApiRequest, res: NextApiResponse) => {
         },
         timeList,
       }: IReqShiftTypeItem = JSON.parse(req.body);
-      const typeIdNew = generateId(
+      const typeIdNew = await generateId(
         `SELECT shift_name_id
         FROM op_shift_name`
       );
--- a/client/src/utils/generateId.ts
+++ b/client/src/utils/generateId.ts
@@ -11,19 +11,10 @@ const checkIsIdExists = async (query: string, id: number) => {
   return Boolean(dbIdFirst);
 };
 
-export const generateId = (query: string) => {
-  let idNew = 0;
-
-  const changeNum = async () => {
+export const generateId = async (query: string): Promise<number> => {
+  let idNew = randomNum();
+  while (await checkIsIdExists(query, idNew)) {
     idNew = randomNum();
-
-    const isIdExists = await checkIsIdExists(query, idNew);
-
-    // if shift name ID exists already
-    // then execute function recursively
-    if (isIdExists) changeNum();
-  };
-  changeNum();
-
+  }
   return idNew;
 };


mbhewitt:
## Adversarial review

Did the 7-point sweep. One **blocker** found; everything else is minor.

### [HIGH] `generateId` will hang forever in 4 of the 9 call sites

The four queries below are passed to `generateId`, but **none of them contain a `WHERE id = ?` clause**. `pool.query(sql, [id])` with no placeholder in `sql` just runs the SQL as-written and silently drops the param. `checkIsIdExists` then asks "did this query return any row?" — and the answer is "yes, the first row of the entire table." So `while (await checkIsIdExists(...))` is `while (true)` whenever the target table is non-empty.

| File:line | Query |
|---|---|
| `client/src/pages/api/shifts/types/index.ts:24-26` | `SELECT shift_times_id FROM op_shift_times` |
| `client/src/pages/api/shifts/types/index.ts:59-61` | `SELECT time_position_id FROM op_shift_time_position` |
| `client/src/pages/api/shifts/types/index.ts:128-130` | `SELECT shift_name_id FROM op_shift_name` |
| `client/src/pages/api/shifts/types/[typeId]/index.ts:511-513` | `SELECT time_position_id FROM op_shift_time_position` |

Before this PR the bug was invisible because `changeNum()` was fire-and-forget — `generateId` returned 0 immediately, the inserts went through with `*_id = 0` (PK collision swallowed silently after the first), and the response shipped before anyone noticed. Bad data, but no hang.

**With this PR**, `await generateId(...)` actually waits, so the first POST against a non-empty `op_shift_name`/`op_shift_times`/`op_shift_time_position` table:
- enters an unbounded SELECT loop slamming the DB,
- pegs one pool connection per concurrent request,
- and (because the call inside `handleTimeListAdd` is buried in `Array.prototype.forEach(async ...)` — see finding 4 below) the outer handler still ships `201 Created` while the connections leak in the background.

The `op_shift_times`/`op_shift_time_position` tables will have rows on any seeded environment (e2e fixture at `client/e2e/helpers/db.ts:194` inserts an `op_shift_name` row in `beforeAll`; prod will obviously have all four populated).

**Fix options (pick one, in this PR):**
1. Add the missing `WHERE id=?` clauses to all four queries (matches the existing pattern in `dates/index.ts`, `roles/index.ts`, `shifts/categories/index.ts`, `shifts/positions/index.ts`, `auth/okta/callback.ts`).
2. Add a max-attempts guard inside `generateId` (e.g. mirror the `for (attempt < 10)` pattern from `client/src/pages/api/volunteers/account/index.ts:20`) — but this only masks the bug; you'd still get an "all attempts collided" throw on the first hit.

Option 1 is the right one. The other 5 call sites already have correct WHERE clauses and will work fine.

### [MED] `handleTimeListAdd` is still fire-and-forget; failures now manifest as response-then-hang instead of response-then-bad-data

`client/src/pages/api/shifts/types/index.ts:149` calls `handleTimeListAdd({ timeList, typeId: typeIdNew })` without `await`. Inside, `timeList.forEach(async ...)` and the nested `positionList.forEach(async ...)` also ignore the returned promises. Even with the WHERE-clause fix above, this means:

- The 201 response can ship before any `op_shift_times` / `op_shift_time_position` row is inserted. A client that immediately fetches the new shift type may see it with no times/positions, then the rows appear seconds later.
- Errors from the inserts (DB down, FK violation, anything) become unhandled rejections — there is no `try/catch` and no caller to receive them. With Node's default `--unhandled-rejections=throw` behavior in newer versions, that's a process-crash risk.

You called this out in the PR body as "out of scope." Confirming that's defensible *for the bug being fixed*, but flagging that the orchestration is now more fragile than it was: previously a `forEach(async)` that throws crashed silently because the bodies were synchronous-with-await-noop after the `generateId(0)` collisions; now the bodies actually do work and actually can throw. Worth a follow-up issue.

The same pattern exists at `client/src/pages/api/shifts/types/[typeId]/index.ts:470,484,532` (three `forEach(async ...)` blocks in `timeListUpdate`/`Add`/`Remove`). Same caveat.

### [LOW] No max-attempts guard in `generateId`

`client/src/utils/generateId.ts:14-20` — `randomNum()` returns in `[1, 1_000_000]`. At ~4k rows in `op_volunteers` the collision rate is ~0.4% per attempt, so this is fine today. But there is no upper bound — a pathological 100% collision (table grew past the random range, DB pinned, network blip, whatever) loops forever. The parallel implementation in `client/src/pages/api/volunteers/account/index.ts:20` uses `for (let attempt = 0; attempt < 10; attempt += 1)` and throws on exhaustion; consider mirroring that here. Low priority — flag for a follow-up.

### [INFO] Call-site count and async context — verified

Confirmed all 9 `await generateId(...)` call sites via `grep -rn "generateId(" client/src`. All 9 are inside `async` functions (Next API route handlers are all `async` here; the two `forEach(async ...)` callbacks are also async). No missed sites, no sync-context type errors. The only other reference to `generateId` in the codebase is a descriptive comment at `client/src/pages/api/volunteers/account/index.ts:16` (that file uses its own `generateNewUserShiftboardId`, not the shared utility).

### [INFO] Numeric correctness — no downstream depended on the `0` sentinel

Traced each call site's use of the returned value: every one feeds directly into an `INSERT ... VALUES (...)` PK column or into the response payload as a created-resource id. No code path treated `0` as "not yet allocated" or similar. The fix is a strict upgrade for the 5 call sites with correct WHERE clauses.

### [INFO] Tests

No unit tests cover `generateId`. The e2e specs at `client/e2e/tests/11-admin-shift-types.spec.ts` only navigate to the create page — they don't actually POST. So neither the original bug nor the HIGH finding above is exercised by CI. The shift-categories, roles, positions, and dates POST endpoints similarly aren't covered by e2e create-flow tests that I could find. **No stale assertions to update**, but also **no regression coverage**, so the HIGH finding will only surface in manual testing / prod.

### [INFO] Race condition (check-then-insert) — pre-existing, not what this PR fixes

The SELECT-then-INSERT in `generateId` plus caller has always been racy: between `checkIsIdExists` returning false and the caller's `INSERT`, another concurrent request can grab the same random id and insert first, causing a duplicate-key error on the second insert. This is unchanged by this PR. Pragmatically fine at current write rates (admin-only endpoints, low concurrency). Worth a one-line comment in `generateId.ts` documenting the assumption, but not a blocker.

---

**Recommendation:** Block on the HIGH finding (add the four missing `WHERE id=?` clauses in the same PR — it's a four-line change and keeps this from being a worse regression than the bug it fixes). MED/LOW are follow-up issues.

mbhewitt:
## Adversarial review

Bottom line: **one new HIGH-severity regression** that this PR turns from a latent no-op into a live hang. The async refactor itself is sound, but it exposes a pre-existing bug in 4 of the 9 callers that was masked when `generateId` always returned `0` synchronously.

### [HIGH] `generateId` will infinite-loop on 4 callers that pass a SELECT with no WHERE clause

`checkIsIdExists` (`client/src/utils/generateId.ts:7-12`) passes `[id]` as bound params, but the following 4 call sites pass a query string with **no `?` placeholder**:

- `client/src/pages/api/shifts/types/index.ts:24-27` — `SELECT shift_times_id FROM op_shift_times`
- `client/src/pages/api/shifts/types/index.ts:59-62` — `SELECT time_position_id FROM op_shift_time_position`
- `client/src/pages/api/shifts/types/index.ts:128-131` — `SELECT shift_name_id FROM op_shift_name`
- `client/src/pages/api/shifts/types/[typeId]/index.ts:511-514` — `SELECT time_position_id FROM op_shift_time_position`

mysql2 silently drops extra params when the SQL has no placeholders (verified locally with `mysql.format(...)`). The query returns **all rows** from the table, so `Boolean(dbIdFirst)` is `true` whenever the table is non-empty, and the new `while (await checkIsIdExists(...))` loop in `generateId` never terminates. The e2e helper at `client/e2e/helpers/db.ts:194-218` INSERTs into all three of these tables, and prod is obviously non-empty — so the very first `POST /api/shifts/types` or update with new positions on any non-empty environment will wedge the request, hold a pool connection, and never respond.

Before this PR the inner `changeNum` recursion would also have looped, but `generateId` returned `0` immediately without awaiting it, so the bug was invisible (the orphan promise just spun in the background and the route returned). Now the route awaits the loop and blocks.

Fixes that would work — pick one:
1. Tighten the call sites to include a `WHERE <id_col>=?` clause (matches the pattern in the other 5 sites).
2. Add a runtime assert in `generateId` that the query contains `?`.
3. Inline a count-based check (`SELECT 1 FROM ... WHERE col=? LIMIT 1`).

Option 1 is the most localized and consistent.

### [MED] `op_shift_name` exhaustion is a real risk; loop is also unbounded

`randomNum()` returns `[1, 1_000_000]`. Even after fixing the HIGH above, the four `op_shift_*` tables have no obvious cap but small population — but `randomNum`'s range collides with the legacy 7-digit shiftboard space, and the loop has no max-attempts guard. The sister implementation `generateNewUserShiftboardId` in `client/src/pages/api/volunteers/account/index.ts:13-32` already does this right (10-attempt cap, throws on failure). Worth porting the same guard into `generateId` so a wedged duplicate-check query or a pathologically dense table can't hang a Next.js worker forever. The `op_volunteers` collision risk you cited (~4k rows, ~0.4%) is the floor; `op_shift_time_position` row counts grow with `shifts x positions` and aren't bounded that low.

### [MED] Pre-existing fire-and-forget in `shifts/types` is now timing-shifted, not made worse

You called this out as out-of-scope, and it stays out-of-scope, but documenting the new behavior:

- `client/src/pages/api/shifts/types/index.ts:149` still calls `handleTimeListAdd(...)` without `await`, and inside `handleTimeListAdd` both `timeList.forEach(async ...)` (line 22) and `positionList.forEach(async ...)` (line 58) ignore returned promises. The response 201s before any INSERTs land — already true before this PR.
- The shift in timing: previously the unawaited promise chain was a near-instant `randomNum()` + a non-awaited check. Now it's a real DB roundtrip per id, so the gap between "response sent" and "row visible" grows from ~ms to tens of ms. Still works in practice but the client-side cache invalidate / refetch on the admin UI is now slightly more likely to race ahead of the writes and show a stale list.
- `client/src/pages/api/shifts/types/[typeId]/index.ts:395` calls `handleTimeListAdd(...)` from inside a `timeListAdd.forEach(async ...)` (line 369) — same fire-and-forget.

Not your bug; file as separate issue per your PR body.

### [LOW] Race on check-then-insert (pre-existing)

Documenting per your ask. Between the SELECT in `checkIsIdExists` and the INSERT in each caller, a parallel request can grab the same random id; whichever INSERT runs second hits the PK duplicate and 500s. Same race the old code had. With `randomNum()` over 1M and tables in the thousands, the collision-on-the-same-tick rate is negligible, but the only correct fix is `INSERT IGNORE` + retry, or an `INSERT ... ON DUPLICATE KEY UPDATE`. Not blocking.

### [INFO] Call-site coverage and async-context audit

- Grep across `client/src` finds exactly 9 `await generateId(` sites and 7 importing files. Matches your count.
- Every enclosing function is `async` (verified: `oktaCallback`, `calendar`, `roles`, `shiftCategories`, `shiftPositions`, `shiftTypeUpdate`, `shiftTypes`, the two anonymous `forEach(async ...)` arrows in `handleTimeListAdd`). `tsconfig.json` has `strict: true`, so any missed site would have failed `tsc`.
- No tests reference `generateId` directly (`client/src/utils/generateId.ts` has no unit test). No assertions in the e2e suite depend on the returned id being `0`. `client/e2e/tests/11-admin-shift-types.spec.ts` only reads the types list, never POSTs — so it would not catch the HIGH-severity infinite loop above.
- Downstream usage of `idNew`: flows into INSERT params and the Okta response payload (`newId` -> `shiftboard_id`). No code path treats `0` as a sentinel post-INSERT; `shiftboard_id = 0` was just a silent collision bomb.

### Recommendation

Do not merge until the HIGH is addressed. Easiest path: add a `WHERE <pk>=?` clause to each of the four bare-SELECT call sites in `shifts/types/index.ts` (3) and `shifts/types/[typeId]/index.ts` (1). The async refactor itself is correct and the loop-vs-recursion choice is fine.

PR #352, opened 2026-05-25
ui(home): host the login affordance on Home, drop the / → /sign-in redirect

mbhewitt:
Reverses the direction of PR #337 + closes #345 partially. Per Chipper + Mew (Discord, 2026-05-25): instead of redirecting unauth users to `/sign-in`, put the login button on Home so the landing page actually says something useful.

## Changes
- **middleware.ts** — `/` back in `PUBLIC_PATHS` so unauth users can reach Home.
- **Home.tsx** — auth-state-aware login block between the "Welcome!" h2 and the welcome body card:
  - **Authenticated** → "Welcome, {playaName} \"worldName\" — view your account" button → `/volunteers/<id>/info`.
  - **Unauth + Okta enabled (off-playa)** → "Sign in to Census" button → `/api/auth/okta`.
  - **Unauth + PIN enabled (on-playa)** → "Sign in with passcode" button → `/sign-in` (the existing PIN form — Chipper explicitly didn't want the volunteer dropdown on Home).
- `/sign-in` stays as a valid route — used by the on-playa PIN flow and as the Okta callback's error-redirect target. Old bookmarks keep working.
- Existing welcome / learn-more body copy unchanged.

## Test plan
- [ ] Off-playa (NEXT_PUBLIC_PIN_ENABLED=false), unauth `GET /` → Home renders with "Sign in to Census" button. Clicking → Okta init.
- [ ] Off-playa, auth `GET /` → Home renders with "Welcome, ... — view your account" button. Clicking → `/info`.
- [ ] On-playa (NEXT_PUBLIC_PIN_ENABLED unset/true), unauth `GET /` → Home renders with "Sign in with passcode" button. Clicking → `/sign-in` with the PIN form.
- [ ] On-playa, auth `GET /` → Home renders with "Welcome, ... — view your account" button.
- [ ] Drawer's "Sign in" link still navigates to `/sign-in` and renders the existing sign-in card.
- [ ] Okta callback error redirects to `/sign-in?error=...` still surface the alert correctly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #390, opened 2026-06-01
fix(session): clear stale client state when census-session cookie is gone

### LINKED ISSUE #389: Stale-client-session causes login spinner that never resolves
From Mew (Discord, 2026-06-01):

> Currently having issues logging in to the server… clicking the login button gives just the spinny thing, never gets to Okta. … It looks like I was half way between logged out and logged in. The home screen still thought I was logged in, but nothing else did, so I think it routed me incorrectly.

## Diagnosis
Client-side `SessionContext` (persisted in localStorage by SessionStorage middleware) outlives the server-side `census-session` HMAC cookie. When the cookie expires/clears but client state still says authenticated:

1. Home renders the "Welcome — view your account" branch (since `isAuthenticated` from context is true).
2. Click → routes to `/volunteers/<id>/info`.
3. Middleware sees no cookie → 307 → `/sign-in?returnTo=/info`.
4. `/sign-in` page reads SessionContext (still "authenticated") and enters a loading/Loading branch waiting on data that requires the cookie — spinner that never resolves.

This UX regression surfaced after PR #352 made Home auth-state-aware. Before, Home didn't switch on auth so the half-state went unnoticed.

## Fix
Single high-level mount validation:
1. New endpoint `GET /api/auth/session` — `withAuth`-wrapped — returns `200 { shiftboardId }` if cookie valid, `401` otherwise.
2. Middleware allowlists `/api/auth/session` so the middleware 401 doesn't preempt the cookie validity check.
3. New hook `useSessionValidation()` runs once at app mount (in `Layout.tsx`). If `SessionContext.isAuthenticated && fetch /api/auth/session === 401`, silently `SESSION_SIGN_OUT` to bring client state in sync. No snackbar / no redirect — just cleans the stale state.
4. After dispatch, next render of the auth-aware components (Home, Header, etc.) picks the unauth branch and the user sees the login button as if they'd never been signed in. Any nav attempt to a gated page still triggers the normal middleware redirect.

## Out of scope
- Periodic re-validation (heartbeat). One-shot on mount is enough for the surfaced bug.
- Migrating SessionContext to a server-source-of-truth pattern. Bigger refactor.

## Test plan
- Reproduce: sign in, then in DevTools delete the `census-session` cookie but leave localStorage intact. Reload Home → should now render the unauth branch (login button), not "Welcome —".
- Cookie still valid → no spurious sign-outs / no extra request noise after the first mount.

mbhewitt:
Closes #389.

## Bug
Surfaced by Mew 2026-06-01 after PR #352 made Home auth-state-aware. When `SessionContext` (persisted in localStorage) outlives the server-side `census-session` cookie, the UI gets stuck in a half-state:
1. Home renders the "Welcome — view your account" branch (client state says authenticated).
2. Click routes to `/volunteers/<id>/info`.
3. Middleware sees no cookie → 307 → `/sign-in?returnTo=/info`.
4. `/sign-in` reads SessionContext (still "authenticated"), enters a loading branch → spinner that never resolves.

## Fix
Three pieces:

1. **New endpoint `GET /api/auth/session`** — `withAuth`-wrapped, returns `200 { shiftboardId }` if cookie valid, `401` otherwise. Cheap probe — no DB hit, withAuth already verified the HMAC.
2. **Middleware** allowlists `/api/auth/session` so middleware's 401 doesn't preempt the cookie-validity check.
3. **New hook `useSessionValidation`** runs once at `Layout.tsx` mount. If client claims authenticated AND probe returns 401, silently `SESSION_SIGN_OUT` to bring client state in sync. No snackbar, no redirect — next render of auth-aware components picks the unauth branch automatically.

One-shot per mount, deliberately not a heartbeat — the bug only manifests when state is already stale at first render; a periodic re-check would open false-positive risk (transient API blips logging users out).

## Test plan
- Reproduce: sign in, then in DevTools delete the `census-session` cookie but leave localStorage intact. Reload Home → renders the unauth branch (login button), not "Welcome —".
- Cookie still valid → no spurious sign-outs, no extra request noise after the first mount.
- Network error to /api/auth/session → user stays signed in (don't bounce healthy users on transient errors).

### DIFF
--- a/client/src/components/layout/Layout.tsx
+++ b/client/src/components/layout/Layout.tsx
@@ -3,6 +3,7 @@ import { ReactNode, useContext } from "react";
 import { DeveloperCard } from "@/components/layout/developer-card";
 import { Footer } from "@/components/layout/Footer";
 import { Header } from "@/components/layout/Header";
+import { useSessionValidation } from "@/hooks/useSessionValidation";
 import { DeveloperModeContext } from "@/state/developer-mode/context";
 
 interface ILayoutProps {
@@ -18,6 +19,10 @@ export const Layout = ({ children }: ILayoutProps) => {
       dateTime: { isEnabled: isDateTimeEnabled },
     },
   } = useContext(DeveloperModeContext);
+
+  // One-shot stale-session detector — runs once per app mount, clears
+  // client state if the cookie has expired (issue #389).
+  useSessionValidation();
 
   // render
   // ------------------------------------------------------------
--- a/client/src/hooks/useSessionValidation.ts
+++ b/client/src/hooks/useSessionValidation.ts
@@ -1,0 +1,53 @@
+"use client";
+
+import { useContext, useEffect } from "react";
+
+import { SESSION_SIGN_OUT } from "@/constants";
+import { SessionContext } from "@/state/session/context";
+
+// Detect a stale client-side session: SessionContext / localStorage says
+// the user is authenticated, but the actual HMAC-signed cookie has
+// expired or been cleared. Without this, the UI keeps rendering the
+// authenticated branch (e.g. Home's "Welcome — view your account"
+// button) and the user gets stuck in a spinner when they click anything
+// that needs the cookie.
+//
+// Approach: on mount, when client state claims authenticated, fire one
+// quiet GET to /api/auth/session. If 401, silently sign out client-side
+// (no snackbar, no redirect) so the next render picks the unauth UI
+// branch. If 200 / network error, do nothing — false positives here
+// would log healthy users out.
+//
+// One-shot per mount; not a heartbeat. The bug only manifests when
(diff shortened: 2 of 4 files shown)


mbhewitt:
LGTM. Handler is GET-only and routes non-GET to 404 as intended (mild nit: 405 is the conventional code, but 404 is fine and matches the brief). Allowlist entry sits cleanly inside the ALLOWLIST array with a useful comment. useSessionValidation correctly bails on !isAuthenticated, only dispatches SESSION_SIGN_OUT on explicit 401, swallows network errors, and uses a cancelled flag in the cleanup; deps array is right. Layout is mounted from app/layout.tsx so the one-shot probe fires per app mount as designed. SESSION_SIGN_OUT constant matches reducer. Playwright CI still pending at review time but no blockers in the code.

PR #392, opened 2026-06-01
feat(mail): collapse add+remove pair when the add hasn't shipped yet

### LINKED ISSUE #391: Email queue: collapse add+remove pair when add hasn't sent yet
From Mew (Discord, 2026-06-01):

> if someone adds themselves to a shift and then removes themselves before the emails get sent, can we mark both as sent (but if the add one has already been sent then the cancel also needs to be sent)

## Goal
When a volunteer churns on a shift (add → remove) before either email actually leaves the queue, both emails should be silently dropped — no inbox noise for a state that resolved itself. But once the "added" email has shipped, the "removed" email must still send (otherwise the volunteer sees a calendar entry they shouldn't).

## Cases
| add state | remove state | desired |
|---|---|---|
| queued | enqueued | both → drop (mark sent or delete) |
| sent | enqueued | leave add, send remove normally |
| sending | enqueued | leave both — race-prone, safer to ship both |

## Implementation sketch
The queue table doesn't currently carry `shiftboard_id` / `time_position_id`, so matching an add to its remove requires either:

1. **Add columns** `shiftboard_id` + `time_position_id` to `op_email_queue`. When the removal enqueue fires, check for queued rows with matching (shiftboard_id, time_position_id, category='assignment', state='queued') and either delete them or flip to a `state='superseded'` so they don't ship. Then suppress the removal enqueue itself.

2. **Add a `dedupe_key` column** (string) — same idea, more generic. Removal's dedupe-key matches the add's. Same lookup + supersede logic.

Either way, gated by per-recipient address match (don't collapse across different `to` addresses).

(1) is more self-documenting; (2) is more reusable for other classes of dedupe later (e.g. multiple admin notifications about the same volunteer's BS sign).

## Out of scope
- The reverse case (remove + readd before send) — could collapse too, but Mew's brief is one-direction.
- Periodic compaction of orphaned queued rows. Separate concern.

## Test plan
- Add yourself to a shift, immediately remove yourself, watch the queue: both rows should land in superseded/deleted state without sending.
- Wait for the add to drain (manual: push tick or wait), then remove: add ships, remove also ships normally.
- Two different volunteers add+remove the same position around the same time: only each one's own pair collapses; no cross-contamination.

mbhewitt:
Closes #391. Per Mew (Discord, 2026-06-01).

When a volunteer adds themselves to a shift and removes themselves again before the assignment email actually drains the queue (rate cap is 100/day, so this happens often), both rows now collapse to `state='sent'` without firing — no inbox noise for a state that resolved itself. Once the assignment has actually shipped, the removal still goes out so the volunteer's calendar drops the event.

## Pieces

1. **migrations/004_email_queue_dedupe_key.sql** — adds `dedupe_key VARCHAR(128) NULL` + `idx_dedupe_state (dedupe_key, state)` to `op_email_queue`. Idempotent (`IF NOT EXISTS`).

2. **client/lib/mail/queue.ts** — new `supersedeQueuedByDedupeKey(pool, key, to)` helper. Flips matching `(dedupe_key, to, state='queued')` rows to `state='sent' / sent_at=NOW() / last_error='superseded by paired enqueue'`. Returns affected row count. `enqueueEmail` now persists the optional `dedupeKey`.

3. **client/lib/mail/index.ts** — re-exports `supersedeQueuedByDedupeKey` so callers don't pass the pool.

4. **client/lib/mail/types.ts** — `EnqueueArgs.dedupeKey?: string`.

5. **client/src/components/api/assignmentNotify.ts** — local `assignmentDedupeKey(shiftboard, timePosition)` helper. Assignment enqueues set it; removal enqueues call `supersedeQueuedByDedupeKey` first and bail if anything collapsed.

## Migration order
Run `004_email_queue_dedupe_key.sql` on prod **before** deploying. The new column appears in the assignment-INSERT column list, so deploying first would error.

```sql
ALTER TABLE op_email_queue
  ADD COLUMN IF NOT EXISTS dedupe_key VARCHAR(128) NULL AFTER category;
ALTER TABLE op_email_queue
  ADD KEY IF NOT EXISTS idx_dedupe_state (dedupe_key, state);
```

## Test plan
- Add self to a shift, immediately remove self → both rows land in state='sent' with `last_error='superseded by paired enqueue'`; no email actually ships.
- Add self, wait for the assignment to drain (worker sends), then remove → removal ships normally (no collapse since no queued add to find).
- Two different volunteers, same time position: each one's add+remove collapses independently — no cross-contamination.
- Existing single-direction assignment emails still send unchanged.

### DIFF
--- a/client/lib/mail/index.ts
+++ b/client/lib/mail/index.ts
@@ -1,6 +1,9 @@
 import { pool } from "../database";
 
-import { enqueueEmail as enqueueEmailRaw } from "./queue";
+import {
+  enqueueEmail as enqueueEmailRaw,
+  supersedeQueuedByDedupeKey as supersedeQueuedByDedupeKeyRaw,
+} from "./queue";
 import { createMysqlStore } from "./store";
 import { selectTransport } from "./transport";
 import type { EnqueueArgs, MailConfig } from "./types";
@@ -43,6 +46,17 @@ export async function enqueueEmail(
   args: EnqueueArgs
 ): Promise<{ id: number; skipped?: boolean }> {
   return enqueueEmailRaw(pool, config(), args);
+}
+
+// Mark queued rows with this (dedupe_key, to) as superseded without
+// sending. Returns affected row count. Callers use this to collapse
+// add+remove pairs that resolved before the add actually shipped —
+// see queue.ts for the contract and #391 for the design.
+export async function supersedeQueuedByDedupeKey(
+  dedupeKey: string,
+  to: string
+): Promise<number> {
+  return supersedeQueuedByDedupeKeyRaw(pool, dedupeKey, to);
 }
 
 // Idempotent: instrumentation.ts calls this on cold start. Safe to call
--- a/client/lib/mail/queue.ts
+++ b/client/lib/mail/queue.ts
@@ -75,8 +75,8 @@ export async function enqueueEmail(
   const [result] = await pool.execute<ResultSetHeader>(
     `INSERT INTO op_email_queue
       (\`to\`, cc, reply_to, \`from\`, subject, body_text, body_html,
-       ics_attachment, ics_filename, category)
-     VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+       ics_attachment, ics_filename, category, dedupe_key)
+     VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
     [
       to,
       joinAddrs(args.cc),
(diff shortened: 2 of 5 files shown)


mbhewitt:
Blocker: the DB is stock MySQL (database/Dockerfile = `FROM mysql`), not MariaDB — `ADD COLUMN IF NOT EXISTS` and `ADD KEY IF NOT EXISTS` are MariaDB extensions and MySQL 8.x rejects them with a syntax error near `IF`. Migration 004 will fail on prod as written. Either drop the `IF NOT EXISTS` (lose idempotency, match the pattern in 001 / 003) or wrap in a procedure that checks information_schema (see migration 003's `ON DUPLICATE KEY` style for the local convention). Other notes are non-blocking: (a) race on `state='sending'` is correctly handled — supersede no-ops and the removal enqueues, which is right since the assignment is mid-flight; (b) migration-before-deploy ordering is called out in the PR body, fine; (c) 57 existing NULL-dedupe_key rows are pre-collapse and don't need backfill; (d) no queue-layer tests in client/tests exist that need updating (mail-worker/backoff/classify don't touch enqueueEmail), but worth adding a unit test for supersedeQueuedByDedupeKey to lock in the (dedupe_key, to, state='queued') match; (e) CI: Playwright still pending. Fix the migration syntax and this is good to merge.

mbhewitt:
Reviewer flagged that `ADD COLUMN/KEY IF NOT EXISTS` is MariaDB-only. Confirmed prod actually runs MariaDB (`mysql from 11.8.6-MariaDB` per the prod client + PROD_DEPLOY.md says "AWS EC2 + RDS MariaDB 11.x"), so the migration syntax is fine as-is. Leaving the IF NOT EXISTS clauses in place for idempotence on re-runs.

PR #398, opened 2026-06-07
fix(shifts): write time_text columns on save + event-day dropdown in time dialog

### LINKED ISSUE #395: Shift Types form: newly added times reappear blank/red — UI writes never set start_time_text/end_time_text
## Symptom (reported by Chipper via Rescue, 2026-06-07)
Rescue added new shifts on the Shift Types form. The times keep disappearing and going back to red as if they still need to be filled out. The shifts DO appear on the Shifts page, but the times are blank/red when reopening the Shift Type.

## Root cause
Read/write column mismatch introduced by the date-ID refactor (f564938, May 13):

- **Read path** — the Shift Types GET (`/api/shifts/types/[typeId]`) was switched to select `st.start_time_text` / `st.end_time_text` + `d.date` (via `start_date_id`) instead of `st.start_time` / `st.end_time`.
- **Write paths** — `handleTimeListAdd` (INSERT, `api/shifts/types/index.ts`) and the `[typeId]` PATCH `timeListUpdate` still write **only** `start_time` / `end_time` (DATETIME) + the date FKs. **Nothing in the app ever writes `start_time_text` / `end_time_text`** — those columns are only populated by the Shiftboard/on-playa ETL.

So any time row created or edited through the UI has `NULL` text columns → the Shift Types form renders empty time fields → red required state. Saving again "fixes" it in the moment (local form state) but the reload comes back blank — hence "keeps disappearing."

The Shifts page is unaffected because `getShiftList.ts` uses `start_time ?? start_time_text` — it prefers the DATETIME column that IS written.

## Evidence
DB sample: ETL rows have `start_time_text='11:30'`; UI-created row `shift_times_id=496824` (`start_time='2025-08-27 08:30'`) has `start_time_text=NULL`, `start_date_id=NULL`.

(Note that row also has a NULL `start_date_id` — worth checking whether the form is sending a bare time or a date outside `op_dates` to the `DATE(?)` FK subquery, which would be a second symptom of the same write-path drift.)

## Fix options
1. **Make writes populate the text columns** (preferred — matches what the ETL produces): set `start_time_text=TIME_FORMAT(?, '%H:%i')` style values (+ `datename`?) in `handleTimeListAdd` and the PATCH UPDATE.
2. Or make the GET tolerant: `COALESCE(start_time_text, TIME_FORMAT(start_time, '%H:%i'))` — but that leaves the columns inconsistent for every other consumer.

Either way, existing UI-created rows need a one-time backfill: `UPDATE op_shift_times SET start_time_text=..., end_time_text=... WHERE start_time_text IS NULL AND start_time IS NOT NULL`.

### LINKED ISSUE #397: Shift Types time dialog: replace free Date picker with event-day dropdown from Dates table
Requested by CaptainMew (Discord, 2026-06-07), alongside the #395 fix.

The time dialog in the Shift Types form currently uses a free MUI DatePicker. Any date not present in `op_dates` produces a NULL `start_date_id`/`end_date_id` (the FK subquery finds no row) and an orphaned shift time.

Replace it with a dropdown (MUI Select) of event days sourced from `/api/dates` (`op_dates`: date + datename), displayed as e.g. "Aug 30 - OpenSun", value = the `YYYY-MM-DD` date. This makes invalid dates unrepresentable and matches how shifts are actually keyed (date IDs) since the date-ID refactor.

Scope:
- `ShiftTypesTimeDialogForm.tsx` — Select fed by `/api/dates`, required validation adjusted (no more "MM/DD/YYYY" placeholder check)
- Read-only time row in `ShiftTypesForm.tsx` — show "MMM DD - datename" too
- No wire-format change: request still sends `startTime`/`endTime` as `"YYYY-MM-DD HH:mm"`

Implementation rides with #395 in one PR since both change the same form/save path.

mbhewitt:
Fixes #395, fixes #397.

## What

**1. Shift Types times no longer come back blank/red (#395)**

Since the date-ID refactor, the Shift Types form GET reads only `start_time_text` / `end_time_text` ("HH:mm") — but the save paths (`handleTimeListAdd` INSERT and the `[typeId]` PATCH `timeListUpdate`) only wrote the DATETIME-shaped `start_time` / `end_time`. Nothing the UI did ever populated the `_text` columns (only the Shiftboard ETL does), so every UI-created/edited time row re-opened as blank/red. The `/shifts` page masked it by falling back to `start_time` (`getShiftList.ts`).

Both save paths now also write the `_text` columns via `DATE_FORMAT(?, '%H:%i')`.

**2. Time dialog: free DatePicker → event-day dropdown (#397)**

The dialog's date field is now a Select fed by `/api/dates` (`op_dates`), labeled "Aug 30 - OpenSun" style. A date that isn't in the Dates table NULLs `start_date_id`/`end_date_id` (the FK subqueries find no row) and orphans the shift time — the dropdown makes that unrepresentable. The read-only time rows in the form show the same "MMM DD - dayname" label. If a stored date is no longer in the Dates table, it still renders as an extra menu entry rather than vanishing.

No wire-format change: the request still sends `startTime`/`endTime` as `"YYYY-MM-DD HH:mm"`, so the API contract and the date-FK subqueries are untouched.

**3. Migration `004_backfill_shift_time_text.sql`**

Idempotent backfill for rows written before the fix (only touches rows where `_text` is NULL/empty). On prod this currently affects the 8 Census Lab Stewardship rows Rescue created (verified: `start_time_text=NULL` on all of them). **Needs a manual run on prod after merge** — same as migrations 001–003.

## Testing

- `npm run build` passes (type-check + all routes compile)
- INSERT and UPDATE SQL shapes executed against a real `op_shift_times` schema in a rolled-back transaction: `_text` columns and both date FKs populate correctly on insert and on date-change update
- Migration executed in a rolled-back transaction against a DB with one known-bad row (`496824`): `still_missing` went 1 → 0, row got `08:30`/`12:30`, rollback clean
- `npm run lint` is currently broken on main (`next lint --fix` — Next 16 dropped the flag), not from this PR

### DIFF
--- a/client/src/app/shifts/types/type/ShiftTypesForm.tsx
+++ b/client/src/app/shifts/types/type/ShiftTypesForm.tsx
@@ -266,8 +266,6 @@ export const ShiftTypesForm = ({
     dateNameMap[dayjs(date).format("YYYY-MM-DD")] = name;
   });
 
-  // logic
-  // ------------------------------------------------------------
   const handlePositionAdd = ({
     alias,
     name,
--- a/client/src/app/shifts/types/type/ShiftTypesForm.tsx
+++ b/client/src/app/shifts/types/type/ShiftTypesForm.tsx
@@ -24,7 +24,7 @@ import {
   TextField,
   Typography,
 } from "@mui/material";
-import { DatePicker, TimePicker } from "@mui/x-date-pickers";
+import { TimePicker } from "@mui/x-date-pickers";
 import { AdapterDayjs } from "@mui/x-date-pickers/AdapterDayjs";
 import { LocalizationProvider } from "@mui/x-date-pickers/LocalizationProvider";
 import dayjs from "dayjs";
@@ -44,6 +44,7 @@ import {
   UseFormSetValue,
   UseFormWatch,
 } from "react-hook-form";
+import useSWR from "swr";
 
 import {
   IFormValues,
@@ -54,12 +55,14 @@ import { ShiftTypesPositionDialogAdd } from "@/app/shifts/types/type/ShiftTypesP
 import { ShiftTypesTimeDialogAdd } from "@/app/shifts/types/type/ShiftTypesTimeDialogAdd";
 import { ShiftTypesTimeDialogUpdate } from "@/app/shifts/types/type/ShiftTypesTimeDialogUpdate";
 import { SnackbarText } from "@/components/general/SnackbarText";
+import { IResDateRowItem } from "@/components/types/dates";
 import type {
   IResShiftTypeCategoryItem,
   IResShiftTypeDefaults,
 } from "@/components/types/shifts/types";
 import { COLOR_BURNING_MAN_BROWN, MEAL_LIST } from "@/constants";
 import { ensure } from "@/utils/ensure";
+import { fetcherGet } from "@/utils/fetcher";
 import { formatDateName, formatTime } from "@/utils/formatDateTime";
 
 dayjs.extend(utc);
@@ -242,9 +245,26 @@ export const ShiftTypesForm = ({
   });
   const [isDialogOpen, setIsDialogOpen] = useState(false);
 
+  // fetcher
+  // ------------------------------------------------------------
+  // event days from the Dates table — used to label saved time rows with
+  // the day name (e.g. "Aug 30 - OpenSun"); the time dialog offers these
+  // same days as a dropdown
+  const { data: dateList }: { data: IResDateRowItem[] | undefined } = useSWR(
+    "/api/dates",
+    fetcherGet
+  );
+
   // other hooks
   // ------------------------------------------------------------
   const { enqueueSnackbar } = useSnackbar();
+
+  // logic
+  // ------------------------------------------------------------
+  const dateNameMap: { [date: string]: string } = {};
+  (dateList ?? []).forEach(({ date, name }) => {
+    dateNameMap[dayjs(date).format("YYYY-MM-DD")] = name;
+  });
 
   // logic
   // ------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


mbhewitt:
## Independent review (subagent)

A separate reviewer agent went through the full diff plus surrounding context. **No correctness blockers or majors found.** Verified specifically:

- **SQL parameter alignment**: INSERT (11 placeholders ↔ 11 params) and UPDATE (11 ↔ 11) both correct; `DATE(?)` / `DATE_FORMAT(?, '%H:%i')` behave correctly on the `"YYYY-MM-DD HH:mm"` strings the form sends.
- **Controlled Select**: menu values and the field value go through the same `dayjs(...).format("YYYY-MM-DD")` normalization, so they always match; the fallback MenuItem covers both the dates-still-loading case and a stored date that was removed from the Dates table (no MUI out-of-range warnings, no vanishing values).
- **Required validation** still fires: the dialog submit handlers check `getValues("timeAdd.date") === ""`, which is exactly the Select's empty state; the removed DatePicker `"MM/DD/YYYY"` onBlur check is unreachable with a Select and not needed.
- **Timezone**: `dateStrings: true` in `lib/database.ts` returns DATE columns as plain strings; `dayjs("YYYY-MM-DD")` parses local with no offset → no day-shift.
- **Why the _text fix is complete**: `ShiftTypesUpdate.tsx` builds `"YYYY-MM-DD HH:mm"` by concatenating the GET's `date` + `_text` time before `reset()` — with NULL `_text` it built `"2026-08-25 "` → invalid dayjs → the blank/red fields in #395. The `_text` write closes that loop.

Two cosmetic nits it raised ("Date is required" copy vs new "Day" label; duplicated `// logic` section header) are fixed in b8f3a0e.

One **minor** noted for the record: in migration 004, a hypothetical malformed `start_time` varchar (not parseable as datetime) would make `DATE_FORMAT` return NULL and leave that row's `_text` NULL with a MySQL warning — not a regression (such a row is already broken), and prod's affected rows are all well-formed.

PR #411, opened 2026-06-13
fix(security): owner-or-admin gate on /api/volunteers/[id]/* endpoints (#410, #350)

### LINKED ISSUE #350: Tighten passcode PATCH auth: any signed-in user can change anyone's passcode

### LINKED ISSUE #410: IDOR: any logged-in user can read ANY volunteer's account info via /api/volunteers/[id]/info
Surfaced by Chipper's point (Discord, 2026-06-12) that "people can only open their own account info." They can't — verified in code.

## The gap
`client/src/pages/api/volunteers/[shiftboardId]/info/index.ts` is wrapped in `withAuth(volunteerInfo)`, so it requires a **valid session** (logged in). But:
- the handler signature is `(req, res)` — it **never receives or uses `session`**,
- it reads `shiftboardId` straight from `req.query` and serves that record,
- there is **no owner check** (`session.shiftboardId === requested`) and **no admin check**.

So **any logged-in volunteer can read any other volunteer's account info** by changing the id in the URL (`/volunteers/<otherId>/info` → `/api/volunteers/<otherId>/info`). That exposes the target's account PII — email, location, arrival, and the other personal fields the account page renders.

This is a classic IDOR. The prod PII hotfix (`d093ae3`) added the *auth gate* (must be logged in) but **not** the *authorization* (must be the owner or an admin). The passcode value itself is separately self-gated (#260), so it's not leaked here — but the rest of the account PII is.

## Severity
P1 / security. Same class as #350 (which is the passcode **write**); this is the broader account **read**. As Chipper noted, exploiting it only takes editing a URL.

## Fix
`withAuth` already passes `session` as the 3rd arg. Have `volunteerInfo` accept it and reject unless owner-or-admin:
```ts
const volunteerInfo = async (req, res, session) => {
  const requestedId = Number(req.query.shiftboardId);
  const isAdmin = await isSessionAdmin(session); // role lookup
  if (session.shiftboardId !== requestedId && !isAdmin) {
    return res.status(403).json({ statusCode: 403, message: "Forbidden" });
  }
  ...
}
```
Audit the sibling endpoints under `/api/volunteers/[shiftboardId]/*` for the same pattern (info/email-preference, other-sap, profile-updated, sap, shifts, account) — any that read/write a volunteer record by URL id need the same owner-or-admin gate.

Related: #350 (passcode PATCH auth), #331 (admin-edit banner).

mbhewitt:
Approved by Mew (Discord, 2026-06-13). Closes the IDOR (#410) + passcode-write hole (#350) and the rest of the same family.

## Problem
`withAuth` guaranteed a valid session (logged in) but the volunteer endpoints then served/accepted whatever `shiftboard_id` was in the URL with **no object-level check**. Any logged-in volunteer could read or act on **anyone's** record by editing the id in the URL.

## Fix
New `client/src/lib/authz.ts` (`isAdmin`, `isOwnerOrAdmin`) + a gate on each endpoint **before any DB access**: a volunteer may act only on their own record; **admins may act on anyone's** (preserves the on-playa admin-helps-volunteer workflow — confirmed with Mew).

Gated: `info` (read PII — #410), `account`, `sap/[sapId]` download, `info/other-sap`, `info/profile-updated`, `shifts`, and the `account/passcode` **PATCH** (#350).

Left intentionally owner-only (unchanged): the passcode **GET** reveal (admins can reset but not read passcodes) and `info/email-preference`.

## Testing
- `npm run build` passes.
- Independent security review (subagent): verdict **safe to deploy** — gates fire before access on all 7, `isOwnerOrAdmin` is parameterized + fails closed (NaN/array/leading-zero coercion can't bypass), passcode GET stays owner-only, full directory coverage, 3-arg signature matches `withAuth`. Summary in a comment below.
- Admin workflow preserved (owner **or** admin).

## Follow-up (separate issue)
The review surfaced a **pre-existing** related gap: the shift PATCH/DELETE sub-handlers (`components/api/shiftVolunteers.ts`) act on a `shiftboardId` from the request **body**, not the URL-gated one — filed separately, out of this PR's scope.

Fixes #410, fixes #350.

### DIFF
--- a/client/src/app/volunteers/[shiftboardId]/info/VolunteerInfo.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/info/VolunteerInfo.tsx
@@ -794,7 +794,10 @@ export const VolunteerInfo = ({ shiftboardId }: IVolunteerInfoProps) => {
         {/* breadcrumbs */}
         <Box sx={{ mb: 3 }}>
           <BreadcrumbsNav>
-            <Link href="/volunteers">Volunteers</Link>
+            {/* Volunteers list is admin-only — don't show the link to
+                non-admins (they'd hit "no permission"). MUI Breadcrumbs
+                drops the falsy child, so no stray separator. */}
+            {isAdmin && <Link href="/volunteers">Volunteers</Link>}
             <Typography>{volunteer.playaName}</Typography>
           </BreadcrumbsNav>
         </Box>
--- a/client/src/lib/authz.ts
+++ b/client/src/lib/authz.ts
@@ -1,0 +1,34 @@
+import type { RowDataPacket } from "mysql2";
+
+import { ROLE_ADMIN_ID, ROLE_SUPER_ADMIN_ID } from "@/constants";
+import { pool } from "lib/database";
+
+// Server-side authorization helpers for the /api/volunteers/[shiftboardId]/*
+// family. `withAuth` already guarantees a valid session (logged in); these add
+// the missing object-level check so a logged-in volunteer can only act on their
+// OWN record, while admins can act on anyone's (the on-playa "admin helps a
+// volunteer" workflow). See #410 (read IDOR) and #350 (passcode write).
+
+// True if the given shiftboard_id holds the Admin or SuperAdmin role.
+export async function isAdmin(shiftboardId: number): Promise<boolean> {
+  const [rows] = await pool.query<RowDataPacket[]>(
+    `SELECT 1
+     FROM op_volunteer_roles
+     WHERE shiftboard_id = ?
+       AND role_id IN (?, ?)
+       AND remove_role = false
+     LIMIT 1`,
+    [shiftboardId, ROLE_ADMIN_ID, ROLE_SUPER_ADMIN_ID]
+  );
+  return rows.length > 0;
+}
+
+// True if the session may act on `requestedShiftboardId`: it's their own
+// record, or they're an admin.
+export async function isOwnerOrAdmin(
+  session: { shiftboardId: number },
+  requestedShiftboardId: number
+): Promise<boolean> {
+  if (session.shiftboardId === requestedShiftboardId) return true;
+  return isAdmin(session.shiftboardId);
+}
--- a/client/src/pages/api/volunteers/[shiftboardId]/account/index.ts
+++ b/client/src/pages/api/volunteers/[shiftboardId]/account/index.ts
@@ -7,9 +7,18 @@ import type {
 } from "@/components/types/volunteers";
 import { pool } from "lib/database";
 import { withAuth } from "@/lib/withAuth";
+import { isOwnerOrAdmin } from "@/lib/authz";
 
-const volunteers = async (req: NextApiRequest, res: NextApiResponse) => {
+const volunteers = async (
+  req: NextApiRequest,
+  res: NextApiResponse,
+  session: { shiftboardId: number }
+) => {
   const { shiftboardId } = req.query;
+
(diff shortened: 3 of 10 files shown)


mbhewitt:
## Independent security review (subagent) — verdict: safe to deploy

Adversarial review of the diff. No SQL injection, no fail-open path, no auth bypass in scope.

- **Gate-before-access**: verified on all 7 — the 403 returns before any `pool.query` (info, account, sap/[sapId], other-sap, profile-updated, shifts, passcode PATCH).
- **`isOwnerOrAdmin`/`isAdmin`**: correct role ids (1/2), `remove_role=false`, parameterized, returns false safely. **Coercion fails closed** — array/garbage `shiftboardId` → `NaN`, which never equals a real id and matches no admin row; leading-zero `"007"` stays consistent between the gate (`Number→7`) and MySQL's coercion of the raw string.
- **Passcode**: GET reveal stays owner-ONLY (no admin path added); PATCH now owner-or-admin, gated before the UPDATE.
- **shifts**: the hoisted `shiftboardId` destructure is correct; top gate covers GET/PATCH/DELETE for the URL id.
- **Coverage**: all 8 URL-id endpoints accounted for (email-preference already self-only by design).
- **3-arg signature**: matches `withAuth`'s call contract; `session` always defined past the 401.

**One MEDIUM, pre-existing, out of scope** → filed as a follow-up: the shift PATCH/DELETE sub-handlers (`shiftVolunteers.ts`) act on a body-supplied `shiftboardId` rather than the URL-gated one. Doesn't block this deploy.

mbhewitt:
Scope note: this PR now also closes two more holes surfaced during review/testing (2026-06-13), both reviewed clean by an independent pass:
- **`/api/volunteers` list endpoint had NO auth at all** (unauthenticated full-roster read) → now `withAuth` + admin-only.
- **`VolunteerInfo` 'Volunteers' breadcrumb** was shown to everyone → gated on admin (a non-admin reached the admin-only list page via it).
So the PR locks down the whole volunteer-endpoint family (per-record owner-or-admin, roster admin-only, passcode reset, and hides the admin-only breadcrumb). The footer menu-visibility inconsistency is tracked separately.

Issue #436, opened 2026-06-16
Art Tour signup rules: require ≥1 work shift, and max 1 Art Tour per volunteer


mbhewitt:
Two front-end signup guards for the Census Art Tour (rider), raised by Chipper 2026-06-16. Both are already stated in the rider description; this makes them enforced.

## 2a — Requires at least one work shift
A volunteer can sign up for the Art Tour **only if they're signed up for at least one other (work) shift this year.** No shifts → can't ride. (Rider desc: *"REQUIRED: You must complete a minimum of one Census work shift while on playa."*)
- Check: the volunteer has ≥1 shift that **isn't** an Art Tour. (Exact "work shift" set may need a definition — at minimum exclude Art Tour itself; possibly exclude perk/social shifts like Welcome Party.)

## 2b — Max one Art Tour
There are two Art Tours, but a volunteer should sign up for **only one** (so there's room for others on the other). Block a 2nd Art Tour signup. (Rider desc: *"please sign up for only ONE Census Art Tour per event."*)
- Check: the volunteer doesn't already have an Art Tour shift.

## Notes
- **Front-end checkable** — the add dialog already has the volunteer's existing shifts (`dataVolunteerShiftList`).
- **Admins override** (consistent with the other signup guards).
- Part of the front-end signup-prevention build alongside #424 (time conflict) and #429 (Camp Steward ≥1 / ≤2).

🤖 Filed by Claude Code on Chipper's request.

mbhewitt:
**Scope clarification (Chipper, 2026-06-16): rider position ONLY.** Both rules (2a require ≥1 work shift, 2b max 1) apply *only* to the **Census Art Tour (rider)** position — the one with **0 CSP**. They do **not** apply to the other Art Tour positions (**Art Tour Lead**, **DataBeast Driver**), which earn CSP and can sign up normally (multiple tours OK).

So:
- The **≥1 work shift** requirement gates only the rider position.
- The **max 1** limit counts only the volunteer's **rider** signups across the two Art Tours (Lead/Driver don't count, and you can lead/drive both tours).

Implementation key: target the specific rider position (`Census Art Tour`, the 0-CSP one), not the Art Tour shift type.

mbhewitt:
[...]

Issue #429, opened 2026-06-15
Checklist item: Camp Steward shift for Census Lab campers


mbhewitt:
New VIP checklist item (Chipper, 2026-06-15 — wanted for this year).

Anyone with the **CensusLabCamp** role (registered as camping at the Census Lab) is expected to sign up for **at least one Census Lab Steward shift**. Add a checklist item:
- **Label:** "Sign up for a Camp Steward shift" → "Camp Steward shift — done".
- **Shown only** when the volunteer has the CensusLabCamp role (role 2000009).
- **Done** once they have a `Census Lab Steward` shift on their schedule; incomplete otherwise.

Data already available in the info endpoint (it fetches roles + shifts): `hasRole(CensusLabCamp) && hasShift('Census Lab Steward')`. Can be a standalone item now; later it'd be a natural rule in the configurable-items model (#279) and a candidate for an inline grid (#266).

🤖 Filed by Claude Code on Chipper's request.

mbhewitt:
**Structural refinement (Chipper, 2026-06-15):** the Camp Steward requirement is a **second requirement on the *same* CensusLabCamp role** — alongside the existing "Meet shift requirements for Census Lab Camp" item (which checks ≥16 CSP). So rather than a *new standalone item*, it likely belongs as a **nested sub-requirement under that existing role item** — the same way the SAP item nests sub-checks (12 CSP + a shift each required day). So: 'Census Lab Camp' item = [≥16 CSP] + [≥1 Camp Steward shift]. Re-scope this as a sub-requirement, not a top-level item.

mbhewitt:
**Add: max-2 block (Chipper, 2026-06-16).** The Camp Steward rule is *one or two* shifts per camper — not more. So alongside the ≥1 checklist nudge, add a signup guard: if a volunteer already has **2 Census Lab Steward shifts**, block/message an attempt to sign up for a **3rd** ('you're only meant to have one or two'). Front-end checkable — the add dialog already has the volunteer's existing shifts (`dataVolunteerShiftList`), so count their Census Lab Steward signups and disable/warn on the 3rd. Low-stakes (catchable in a report), but automating it is the ask. Same Census Lab Steward feature as the ≥1 nudge above — implement together.

Issue #279, opened 2026-05-03
Architecture: Tenant-configurable strings, checklist items, and eligibility rules


mbhewitt:
# Architecture: Tenant-configurable strings, checklist items, and eligibility rules

## Background

The app is intended to be usable by other Burning Man departments (PEIRS, Placement) and potentially other volunteer-run events — see #218 for the umbrella multi-tenant goal. Today, most user-visible behavior is Census-specific and hardcoded:

- **UI strings** — labels, banner copy, form prompts, button text are baked into JSX
- **Checklist items** — what items appear, what completes them, what links they point to live as React component code
- **Eligibility rule parameters** — CSP thresholds, day-by-day shift requirements, role-based bypasses are inline in the SAP eligibility engine

Other tenants would need to fork to change any of this. That's the wrong default for a multi-tenant platform.

#237 already covers the roles/permissions slice of the same problem. This issue is about everything else — strings, checklist content, and rule parameters.

## Out of scope

- Full theming / branding / colors (separate concern)
- Per-user overrides — this is per-tenant only
- Migrating existing data (separate effort, after design lands)
- The roles/permissions config layer — owned by #237

## Requirements

A design exercise that produces:

1. **A taxonomy** — for the VIP page (and account / sign-in / shifts pages by extension), classify every user-visible decision as either:
   - **Structural** — applies to all tenants (e.g., "the page has a checklist section")
   - **Tenant-configurable** — Census's value is the default but each tenant overrides (e.g., the literal label text "Tell us about where you will be camping?")

2. **A storage proposal** — where do tenant-configurable values live?
   - Config file (JSON/YAML in repo)?
   - Database table (e.g., `op_tenant_config` with key/value pairs)?
   - Hybrid?
   - How are translations / wording variants of the same field handled?

3. **A checklist-item model** — checklist items today are React components with conditional rendering. Proposal for representing them as data:
   - What's the schema for a checklist item (label, link, completion-trigger, sub-items)?
   - How does completion-trigger work as data (role check? boolean field on volunteer? composite expression)?
   - How do tenants add/remove items without code changes?

4. **An eligibility-rule model** — the SAP CSP threshold, day-by-day requirements, role-based bypass thresholds. Proposal for these as configurable parameters rather than hardcoded.

5. **A migration strategy** — how do we evolve existing hardcoded sites toward this model without breaking Census's flow? Phased approach is fine.

## Deliverable

[...]

PR #438, opened 2026-06-17
Config-driven signup-rule warnings (art tour / stewardship)

mbhewitt:
Adds two **generic, DB-backed** signup-rule knobs so these rules are *configuration*, not hardcoded per-shift UI logic (per Mew 2026-06-17 — addresses the "I feel weird about single-use config in the UI" concern, and is a step toward #279). Enforcement is a **front-end warning that allows proceed** — admins unaffected, no hard block (Mew's call over a hard fail).

Addresses **#436** (art tour: ride only one, must be active) and **#429** (stewardship: 1–2 not more) — as warnings.

## How it works
Two nullable columns on `op_position_type` (migration 006):
- **`max_per_volunteer`** — most signups one volunteer should hold for this position across all shifts. Census Art Tour = **1**, Census Lab Steward = **2**.
- **`min_scheduled_csp`** — CSP a volunteer must already have scheduled to take it. Census Art Tour = **3** (the no-CSP rider seat is for active volunteers; smallest work shift = 3 CSP).

The add/signup dialog reads them and, when the selected position carries a rule, shows a **warning snackbar** (mirroring the existing time-conflict warning) if the candidate would exceed the cap or fall short on CSP. It never blocks; the next rule like this is a **data edit, not a code change**.

The two art tours are one shift type ("Art Tour", two times) sharing the open "Census Art Tour" rider position, so "one tour max" is just `max_per_volunteer = 1` on that position — no special grouping.

## Files
- `migrations/006_*.sql` — columns + values
- `GET /api/shifts/[timeId]/volunteers` — returns the two knobs per position
- `GET /api/volunteers/[id]/shifts` — also returns `positionId` + `csp` per signup (so the dialog can count a position type & total scheduled CSP)
- `ShiftVolunteersDialogAdd.tsx` — the warning effect
- types + `scheduler_schema.sql`

## Verification
- `tsc --noEmit`: clean
- `next build`: clean (all routes compile)
- migration applied to local DB OK

## Deploy note
Migration 006 must run on prod **before/with** this code (adds the columns). `Census Lab Steward` isn't in the local subset, so its UPDATE matched 0 rows locally — it **will** match on prod.

## Test on a build before prod
- [ ] Select the "Census Art Tour" rider for a volunteer already on the other tour → warning, can still proceed
- [ ] Select it for a volunteer with < 3 scheduled CSP → warning, can still proceed
- [ ] Volunteer with ≥ 3 CSP & no existing tour → no warning
- [ ] Census Lab Steward: warn on the 3rd steward signup
- [ ] Admin sees the warning but is never blocked


mbhewitt:
🤖 Automated diff review (independent agent)

Reviewed `feat/signup-rule-config-warnings` vs `origin/main`. `npx tsc --noEmit` passes clean. **No blockers** — the change is correct and consistent with existing patterns. Detail below.

### ✅ Verified correct

**1. Hook ordering.** `timePositionIdShiftWatch` hoisted to `ShiftVolunteersDialogAdd.tsx:133` (next to the other `watch()` calls); the old post-early-return declaration is removed (the line at the former `:300` is gone). The new `useEffect` sits at `:275-336`, before the early returns (`:340`, `:356`). No rules-of-hooks violation, no duplicate declaration. Deps array is complete.

**2a. max_per_volunteer.** `:294-298` counts active signups where `shift.positionId === positionId` and excludes the slot being added via `shift.timePositionId !== timePositionIdShiftWatch`, warns at `heldCount >= maxPerVolunteer`. For Art Tour (max 1): if the volunteer holds the *other* tour (different `timePositionId`, same `positionId`), `heldCount = 1 >= 1` → warns. Correct "ride only one of two" behavior. The exclusion correctly prevents a re-add of the same slot from self-counting.

**2b. min_scheduled_csp.** `:314-317` sums `shift.csp ?? 0` over active shifts, warns at `< minScheduledCsp`. Sum logic is correct. One semantic note (not a bug): the total includes *all* the volunteer's active signups regardless of position, including the rider seat itself if already held — which is the intended "do they have enough work scheduled" reading. Just confirm Art Tour's own `sap_points` is 0 in the data, otherwise a volunteer holding only the tour seat could self-satisfy the 3-CSP floor on re-evaluation.

**3. Data plumbing.** Both APIs check out:
- `/api/shifts/[timeId]/volunteers` selects `pt.max_per_volunteer, pt.min_scheduled_csp` and maps null-safe to `maxPerVolunteer`/`minScheduledCsp` (index.ts:41-42, 118-120). `IResShiftPositionCountItem` updated with `number | null` (types/shifts/index.ts:41-42).
- `/api/volunteers/[id]/shifts` selects `stp.position_type_id, stp.sap_points` → `shift.positionId`/`shift.csp` (index.ts:34-35, 81/86, 90/95). Destructured names match the SELECT aliases. `IResVolunteerShiftItem.shift` gains `csp: number` + `positionId: number` (types/volunteers/index.ts:61,65). Types line up; tsc confirms.

**4. Migration 006.** `ALTER ... ADD COLUMN ... AFTER role_id` + two UPDATEs. Matches 001's run-once plain-ALTER style. `scheduler_schema.sql:84-85` updated to match (fresh DBs get the columns). No FK/charset concerns — plain nullable INTs. UPDATEs are idempotent (header comment is accurate); the ALTER is **not** idempotent (also accurately stated). Re-running the file errors on `Duplicate column` — fine given the documented run-once convention.

### ⚠️ Minor / non-blocking

- **Snackbar can stack on re-render.** `SnackbarProvider` (layout.tsx:238) has no `preventDuplicate`, and the effect re-fires whenever `timePositionIdShiftWatch`/`volunteerSelected`/`dataVolunteerShiftList` change. Re-selecting the same position re-queues the warning. This is **identical to the pre-existing time-conflict effect** (:219-268), so it's not a regression — flagging for parity awareness only.
- **`volunteerSelected` can be falsy in the label.** `:284` builds `volunteerLabel` from `volunteerSelected?.playaName`/`worldName`; if undefined it renders `undefined "undefined"`. In practice the effect only runs meaningfully once `timePositionIdShiftWatch` is set, which requires a volunteer to be picked first, and `dataVolunteerShiftList` is keyed on the selected volunteer — so the race window is tiny. The existing time-conflict warning has the same `volunteerSelected?.` exposure, so again no regression. Could tighten by early-returning when `!volunteerSelected`.

### Other checks
- No existing e2e test references these positions/columns or the add-dialog warning text; nothing should break (admin-volunteer-management spec doesn't assert on snackbars here).
- `remove_shift=false` already filters at SQL level; the effect's extra `!shift.canceled` filter is harmless belt-and-suspenders.

mbhewitt:
Superseded by #445, which includes this PR's config columns/migration plus the agreed signup blocks + Camp Steward checklist. Merged & deploying #445.

PR #442, opened 2026-06-22
fix(volunteers): exclude removed shifts from Attended/Absent/Remaining counts; spell out column labels

mbhewitt:
## Bug
The admin **Volunteers** list counts each volunteer's Attended / Absent / Remaining shifts by `LEFT JOIN`ing `op_volunteer_shifts` **without filtering `remove_shift`** — so dropped/removed shifts are counted too. Reported by Chipper: Rescue showed **11 remaining** but has **8** (3 were removed shifts).

**Fix:** add `AND vs.remove_shift=false` to the join (kept in the `ON` clause so the `LEFT JOIN` still includes volunteers with zero shifts). Verified against prod: Rescue has 8 active `noshow='X'` rows + 3 removed → the filter yields the correct 8.

## Also (per Chipper)
- Spell out the abbreviated headers: **Att./Abs./Rem. → Attended/Absent/Remaining**. Safe — the table data is array-of-arrays (columns map by position, not by `name`).
- Remove the now-redundant **"Legend"** section (it explained the abbreviations) + its now-unused `Card`/`CardContent`/`Grid` imports.

## Verification
- `tsc --noEmit`: clean
- `next build`: clean
- Genuine bug → fits the launch-freeze "bugs are fine" exception.

### DIFF
--- a/client/src/app/volunteers/Volunteers.tsx
+++ b/client/src/app/volunteers/Volunteers.tsx
@@ -1,15 +1,7 @@
 "use client";
 
 import { Chat as ChatIcon } from "@mui/icons-material";
-import {
-  Box,
-  Card,
-  CardContent,
-  Chip,
-  Container,
-  Grid,
-  Typography,
-} from "@mui/material";
+import { Box, Chip, Container, Typography } from "@mui/material";
 import { green, grey, red } from "@mui/material/colors";
 import { FilterType } from "mui-datatables";
 import { useRouter } from "next/navigation";
@@ -83,7 +75,7 @@ export const Volunteers = () => {
       },
     },
     {
-      name: "Att.",
+      name: "Attended",
       options: {
         filter: false,
         setCellHeaderProps: setCellHeaderPropsCenter,
@@ -93,7 +85,7 @@ export const Volunteers = () => {
       },
     },
     {
-      name: "Abs.",
+      name: "Absent",
       options: {
         customFilterListOptions: {
           render: (value: string) => `Absent: ${value}`,
@@ -117,7 +109,7 @@ export const Volunteers = () => {
       },
     },
     {
-      name: "Rem.",
+      name: "Remaining",
       options: {
         customFilterListOptions: {
           render: (value: string) => `Remaining: ${value}`,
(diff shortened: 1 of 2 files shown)


mbhewitt:
🤖 Automated diff review (independent agent)

Typecheck: `npx tsc --noEmit` passes clean on this branch (exit 0).

✅ **SQL fix is correct.** `AND vs.remove_shift=false` lives in the `ON` clause, so the `LEFT JOIN` is preserved: volunteers whose shifts are *all* removed (or who have none) still return one row with `noshow=NULL`, which falls through the `default` case and yields a 0/0/0 row instead of vanishing from the roster. `remove_shift` is `tinyint(1) DEFAULT '0'` (`database/scheduler_schema.sql:262`), and `vs.remove_shift=false` is the exact idiom already used in `shifts/index.ts:49/91`, `shifts/[timeId]/volunteers/index.ts:93`, etc. Consistent and right.

✅ **Header rename is positional-safe.** `dataTable` is an array-of-arrays (`Volunteers.tsx:183-203`), so column `name` is a display label only, not a data key. "Att./Abs./Rem." → "Attended/Absent/Remaining" changes nothing functionally. The cell→column mapping is intact (noShowCount→Absent/red chip, remainingCount→Remaining/green chip).

✅ **No leftovers.** Legend `<Box>` fully removed; no remaining references to it. Dropped imports `Card`/`CardContent`/`Grid` are no longer used; `Box`/`Typography` are still used (`Volunteers.tsx:188,236,242`). Filter render strings ("Absent: ", "Remaining: ", `:91/:115`) are unchanged and still fine.

⚠️ **Partial reconciliation with the detail view (not a blocker).** The per-volunteer detail page (`volunteers/[shiftboardId]/info/index.ts:194-197`) excludes shifts on THREE conditions: `vs.remove_shift=false` AND `stp.remove_time_position=false` AND `st.remove_shift_time=false`. This PR only adds `vs.remove_shift=false` to the roster count. So a signup whose shift-*time* or time-*position* was removed (while `vs.remove_shift` is still 0) will still be counted on the roster but won't appear in the detail view — a residual mismatch. It fixes the common/stated case (dropped signups); flagging the edge case for a possible follow-up, not gating this PR.

**Blockers: none.**

Issue #444, opened 2026-06-22
Configurable signup-rules engine (warning / block / Discord ping / checklist + custom copy)


mbhewitt:
**Vision (Chipper, 2026-06-22):** a single admin-configurable engine for signup rules, so we stop hard-coding each one.

Today, each signup constraint is a bespoke code change. This issue proposes a unified **rules model** where an admin can, per rule:

- **Define the trigger** — e.g. double-booking (overlapping shift times), two of the same special shift (art tour), a prerequisite/CSP minimum, a fatigue combo (early + late on adjacent days), camping-role expectations.
- **Choose the action** — `warning` (allow, advise) · `block` (hard-stop, non-admin) · `discord-ping` (notify a leader, e.g. Rescue) · `checklist-item` (show a to-do on the account page). Actions aren't mutually exclusive (a rule could warn the volunteer *and* ping a leader).
- **Choose the audience** — who it affects (volunteer vs admin-exempt) and who gets notified (which leader / Discord target).
- **Define the copy** — the exact front-end wording shown to the volunteer when it fires, editable without a deploy.

**Two independent axes** worth noting in the design: *action type* (warn/block/ping/checklist) and *audience* (who's blocked vs who's notified). The fatigue combo is "allow the volunteer, ping Rescue"; double-booking is "block the volunteer" — the model needs both axes.

**Subsumes these current one-offs:** #424 (double-book block), #438 (double art-tour block), #436 (art-tour ≥1 CSP gate), #429 (camp-steward checklist), the eligibility filter, and the 4am+6:30pm fatigue Discord ping.

**Related:** #279 (configurable-items model) — this is the natural home / extension of that work.

Scope: **next year** (not launch). Filed on Chipper's request to log the direction.

🤖 Filed by Claude Code on Chipper's request.

Issue #424, opened 2026-06-15
Shift signup allows double-booking: time-conflict check is warning-only (and detection is partial)


mbhewitt:
## Problem
A volunteer can be signed up for two **overlapping** shifts. Surfaced by Woodie (and re-confirmed 2026-06-14 via a signup audit):

> Wingnut Woodie, **Aug 26**: Setup/Setup Crew **09:00–13:00** *and* Gate Sampling/Gate Sampler **11:30–15:30** → overlap 11:30–13:00.

Chipper's framing: *"the fact that Woodie double-booked is less interesting than the fact that Woodie **could** double-book."*

## Current behavior
`client/src/app/shifts/[timeId]/volunteers/ShiftVolunteersDialogAdd.tsx` already detects a conflict and shows a **warning snackbar** (`variant: "warning"`, *"...shift will cause a time conflict for..."*) — but it is **advisory only**. The add still goes through; nothing blocks the submit. So a warning flashes and the double-booking is created anyway.

## Two gaps
1. **Not enforced.** The conflict warning doesn't prevent the signup. Decide the policy: hard-block, or require an explicit *"this overlaps X — sign up anyway?"* confirmation. (Admins may need an intentional override; self-signup probably shouldn't double-book silently.)
2. **Detection is partial.** It only checks whether the *new* shift's **start** falls inside an existing shift (`dayjs(startTime).isBetween(existing.start, existing.end)`). It misses overlaps where the new shift starts *before* and runs into an existing one, or fully contains it. Should be a true range-intersection test (`newStart < existingEnd && newEnd > existingStart`).

## Acceptance
- Adding an overlapping shift is **blocked or requires explicit confirmation** (per chosen policy).
- Overlap detection catches **all** true time-range intersections, not just start-inside.
- Server-side enforcement too (the POST add handler currently doesn't check conflicts at all — UI-only checks can be bypassed). 

🤖 Filed by Claude Code on Chipper's request.

mbhewitt:
**Refinement (Mew, 2026-06-15):** the policy is **gate normal volunteers, allow admin override.** Woodie could double-book *because he's an admin* — that's intended. Normal (non-admin) volunteers should be **blocked** from creating an overlapping signup; admins keep the ability to assign overlapping shifts deliberately.

[...]

PR #445, opened 2026-06-22
feat(signup): block ineligible signups + Camp Steward checklist (#424, #438, #436, #429)

mbhewitt:
Bundles the launch-priority signup rules Rescue & Chipper asked for (2026-06-22). **Supersedes #438** — it includes #438's config columns/warnings and builds on them. Deploy is gated on Mew (launch front-end freeze); this PR is for review + testing.

## What it does

**Enforcement model** (per Chipper/Rescue): on **Add (submit)**, a non-admin is **hard-stopped** — a red error snackbar pops and the signup does **not** happen. Admins are exempt: they keep the orange warning on selection and can override. Client-side gate, consistent with the existing full-shift/slot gating; no server change.

- **#424 — double-booking.** Blocks signing up for a shift that overlaps one you already hold. Strict overlap, so legit back-to-back shifts (touching endpoints) are still allowed.
- **#438 — over-limit positions.** Blocks exceeding a position's `max_per_volunteer` (e.g. a 2nd Census Art Tour; Census Lab Steward capped at 2).
- **#436 — art-tour CSP gate.** Blocks taking the Census Art Tour seat without at least 1 scheduled CSP (a real work shift). Migration sets `min_scheduled_csp` **3 → 1** (0-CSP shifts like Welcome Party / DataBash don't count).
- **#429 — Camp Steward checklist item.** "Sign up for one or two Camp Steward shifts" on the account checklist, shown only to Census Lab campers (`CensusLabCamp` role), marked done once they hold a `Census Lab Steward` shift.

## Deploy notes
- Run **`migrations/006_add_signup_rule_columns.sql`** (adds the two columns + seeds Art Tour / Lab Steward). Run-once plain ALTER.
- `NEXT_PUBLIC_*` not affected.

## Not in this PR
- The read-only **Discord double-book / fatigue notifications to Rescue** are a standalone script (no app change), tracked separately.
- Future: collapses into the configurable-rules engine, #444.

## Testing
- Build + typecheck green locally.
- Manual: as a non-admin, try to (a) double-book an overlapping shift, (b) take a 2nd art tour, (c) take the art tour with 0 CSP → each pops a red error and is blocked. As an admin, each only warns and proceeds. Census Lab camper sees the checklist item; it completes after adding a Lab Steward shift.


mbhewitt:
**Verdict: LGTM with nits.** Reviewed the diff adversarially — the core block logic (overlap math, admin exemption, max-per-volunteer counting, CSP gate) is correct, in-scope, and parses dates consistently. Typecheck passes clean. No correctness bug that would wrongly block a legitimate non-admin signup. A few low-severity gaps below.

**Verified correct**
- **#424 overlap math** (`ShiftVolunteersDialogAdd.tsx:613-617`): `dayjs(shift.startTime).isBefore(dayjs(endTime)) && dayjs(startTime).isBefore(dayjs(shift.endTime))` is a correct strict half-open overlap. Back-to-back A=[10,11]/B=[11,12] → `11<11` is false → no conflict (allowed). True overlap fires. `startTime`/`endTime` are the props of the shift being added; `shift.*` is the held shift — variables correct.
- **Date parsing consistency**: both `dataShiftVolunteersItem` (props) and `dataVolunteerShiftList` format times as `start_time ?? start_time_text`, and `op_shift_times.start_time` is a full `"YYYY-MM-DD HH:mm"` datetime (per e2e seed `18-vip-page.spec.ts:43-50`), so dayjs parses both sides identically and overlap includes the date — no cross-day false positive, no tz drift.
- **Admin exemption** (`:606`): `if (!isAdmin && dataVolunteerShiftList)`. `isAdmin` (`:368`) is defined above `onSubmit` (`:601`) in the same component body, both reached only after the loading guards — in scope. Keys off the operator's session role, matching "admins keep orange warnings, non-admins blocked."
- **#438 max-per-volunteer** (`:626-633`): counts active shifts with matching `positionId`, excludes the slot being added — correct. `positionId` is `op_position_type.position_type_id` on both sides.
- **#436 CSP gate** (`:644-648`): sums `shift.csp ?? 0` over active shifts; null-safe; Art-Tour seat isn't yet held so it can't satisfy its own gate.
- **#429 checklist** (`VolunteerInfo.tsx:779-781`): `roles.includes("CensusLabCamp")` matches existing internal-id usage (`:470-471` use `"OtherSAP"`/`"Staff"`); `shifts.some(s => s.position === "Census Lab Steward")` matches the typed shape (`volunteer-info.ts:69-74`). `shifts` is correctly destructured (`:461`).
- **Migration 006**: numbering correct (next after 005); ALTER run-once + idempotent UPDATEs as documented; end-state `Census Art Tour` min_scheduled_csp=1 / max=1, `Census Lab Steward` max=2 matches the issues. No "3" appears — the "3→1" in the PR description is just describing the target value; final value is 1. Fine.
- **No UX dead-end** (#7): the block `return`s before `trigger`, so `isMutating` stays false and the submit button (`:1130`, disabled only on `errors`/`isMutating`) stays enabled — user reads the red snackbar and can re-select/re-submit.

**Nits / low-severity**
- **[low] Fail-open if volunteer shifts not loaded** (`:606`): `if (!isAdmin && dataVolunteerShiftList)` — when `dataVolunteerShiftList` is `undefined` (SWR still in-flight or errored), the entire block is skipped and a non-admin submit proceeds unchecked. Narrow (auto-selected self-signup usually has it loaded), but consider gating submit on the fetch being settled, or treating undefined as "can't verify → don't allow" for non-admins.
- **[low] Client-only enforcement**: server is unchanged (acknowledged in the PR). A non-admin hitting `POST /api/shifts/[timeId]/volunteers` directly bypasses all four blocks. Acceptable per the stated scope, but it means these are guard-rails, not a security boundary — worth a follow-up issue if any of these are meant to be hard rules.
- **[low] Canceled Lab Steward shift still marks #429 done**: the info-page shifts query (`info/index.ts:182-200`) filters `remove_shift=false` but not `st.canceled`, so a canceled-but-not-removed Census Lab Steward shift would still flip the checklist to "done." Edge case; matches how the rest of that list behaves.
- **[nit] Warning-effect label when volunteer not yet resolved** (`:284`): `volunteerLabel` reads `volunteerSelected?.playaName` but the effect only guards on `dataVolunteerShiftList`/`timePositionIdShiftWatch`, not `volunteerSelected`. If `dataVolunteerList` lags, the snackbar can render `undefined "undefined"`. Cosmetic; add `volunteerSelected` to the early-return guard.

Note unrelated to this PR: `npm run lint` is broken on the branch (`next lint --fix` → "unknown option '--fix'", Next 16 dropped `next lint`). `tsc --noEmit` passes clean.

PR #453, opened 2026-06-23
feat(auth): roles-on-login roster — auto-grant camp/staff roles at login

mbhewitt:
Builds the "roles on first login" facility Mew asked for (2026-06-22).

## What it does
- **Migration 007** adds `op_role_grant_roster` (email → role_id), seeded from the authoritative **Lab / Counter Culture / Staff** roster (47 pairs).
- The **Okta login callback** now, after resolving the account, grants any roster roles for the volunteer's email they don't already hold — **additive only, never removes** — wrapped in try/catch so a roster miss or a table-less deploy never blocks login.

## Why
Camp-roster folks who've never logged into the app can't be given roles ahead of time — `op_volunteer_roles` has an FK to `op_volunteers`, so the row must exist first. This grants their role the moment they first sign in. Same spirit as the sampler-role recommendation in census-python, but keyed to a manual roster instead of shift history.

## Deploy
- Run **`migrations/007_add_role_grant_roster.sql`** (creates + seeds the table; idempotent).
- No env changes.

## Adding people later
`INSERT IGNORE INTO op_role_grant_roster (email, role_id) VALUES (LOWER('…'), 2000009|2000008|2000006);` — applies on that person's next login.

## Testing
- Build + typecheck green. Grant is additive + idempotent (`ON DUPLICATE KEY UPDATE add_role=true, remove_role=false`); the removed wrong-Lab roles won't return since those people aren't on the Lab roster.

### DIFF
--- a/client/src/pages/api/auth/okta/callback.ts
+++ b/client/src/pages/api/auth/okta/callback.ts
@@ -394,6 +394,34 @@ const oktaCallback = async (req: NextApiRequest, res: NextApiResponse) => {
       shiftboardId = newId;
     }
 
+    // Roles-on-login: grant any camp/staff roles the roster (migration 007)
+    // lists for this email that the volunteer doesn't already hold. ADDITIVE
+    // only — never removes a role. This is how camp-roster folks who've never
+    // logged into the app get their role the moment they first sign in (they
+    // can't have op_volunteer_roles rows before their account exists, FK). Same
+    // spirit as the sampler-role recommendation in census-python, but keyed to
+    // a manual roster instead of shift history. Wrapped so a roster miss or a
+    // deployment without the table never blocks login.
+    try {
+      const [rosterRows] = await pool.query<RowDataPacket[]>(
+        "SELECT role_id FROM op_role_grant_roster WHERE LOWER(email)=LOWER(?)",
+        [email]
+      );
+      for (const { role_id } of rosterRows) {
+        await pool.query(
+          `INSERT INTO op_volunteer_roles (shiftboard_id, role_id, add_role, remove_role)
+           VALUES (?, ?, true, false)
+           ON DUPLICATE KEY UPDATE add_role=true, remove_role=false`,
+          [shiftboardId, role_id]
+        );
+      }
+    } catch (rosterErr) {
+      console.warn(
+        "roster role grant skipped:",
+        rosterErr instanceof Error ? rosterErr.message : rosterErr
+      );
+    }
+
     // build the account response (same shape as passcode sign-in)
     const account = await buildAccountResponse(shiftboardId);
     if (!account) {
--- a/migrations/007_add_role_grant_roster.sql
+++ b/migrations/007_add_role_grant_roster.sql
@@ -1,0 +1,70 @@
+-- Migration: roles-on-login roster (per Mew 2026-06-22).
+--
+-- A small authoritative roster of camp/staff roles keyed by email. On login,
+-- the Okta callback looks the volunteer up by email and grants any roles listed
+-- here that they don't already hold (ADDITIVE only — it never removes a role).
+-- This is the "roles on first login" facility for the camp roster (Census Lab,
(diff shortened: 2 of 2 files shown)


mbhewitt:
**Independent review — verdict: LGTM-with-nits.** The change is correct, additive-only, and cannot block login. No privilege-escalation hole given the BM Okta trust model. Findings below; none are blocking.

### Correctness — PASS
- Grant block sits after all four `shiftboardId` resolution branches (okta_id / email / sb_pinfo / create-new), so `shiftboardId` is guaranteed non-null at grant time. ✅
- FK `op_volunteer_roles.shiftboard_id → op_volunteers.shiftboard_id` holds: every branch either found or just-INSERTed the op_volunteers row before reaching the grant. ✅
- `op_volunteer_roles` PK is `(shiftboard_id, role_id)`, so `ON DUPLICATE KEY UPDATE add_role=true, remove_role=false` is exactly the right idempotent upsert. Re-login is a harmless no-op for already-held roles. ✅
- Seed is lowercased and the lookup is `LOWER(email)=LOWER(?)`, so case can't cause a miss. ✅

### Can it block/break login? — PASS
- Grant is wrapped in its own try/catch; a missing `op_role_grant_roster` table, a DB error, or an FK violation just `console.warn`s and continues. Login still succeeds. ✅
- It's also nested inside the outer try, so even an unforeseen throw can't escape to crash the handler. ✅

### Security / privilege escalation — PASS (one nit)
- Roles are keyed to `userInfo.email` fetched **server-to-server** from Okta `/v1/userinfo` with the access token — not a client-supplied value. A user cannot spoof another person's email to claim their roster role; the email reflects whoever actually authenticated against the BM Okta tenant (`login.burningman.org`). Granting Staff (2000006) on email-match is acceptable because the roster is admin-controlled and the email is IdP-asserted. ✅
- **NIT (low):** the code trusts `userInfo.email` without checking an `email_verified` claim. For BM's single corporate Okta tenant this is fine (emails aren't self-asserted there), but if the tenant ever federates a social/external IdP that allows unverified emails, this becomes a soft escalation vector. Cheap hardening: assert `userInfo.email_verified === true` before the grant (and arguably before account creation generally). Not blocking.

### Idempotency / removed-people — PASS
- Verified the four people deliberately removed from CensusLabCamp are NOT granted role 2000009 in the seed: `mu@burningman.org` (has 2000006/2000008 only), `chipper@burningman.org` (2000006), `ann976norton@gmail.com` (2000006/2000008), `jimstamper@mail.com` (2000008). None get the Lab role, so re-login won't resurrect a deliberately-removed grant. ✅

### SQL injection — PASS
- Both new queries use parameter binding (`?`) for email and shiftboardId/role_id. `role_id` is read straight from the table (BIGINT column), never interpolated. No injection surface. ✅

### Additional notes (non-blocking)
1. **FK to op_roles + per-row failure semantics (low):** roles 2000008/2000009 (CounterCulture/CensusLab camp) are not seeded by any in-repo migration — they must already exist in prod `op_roles` (2000006 Staff does, per migration 003). If a target DB lacks one of those role_ids, that row's INSERT throws an FK error, which (because the loop is inside one try) aborts the loop and **skips the remaining roster rows for that user**. So a user with [2000006, 2000009] could get 2000006 granted but silently miss 2000009 if 2000009 is absent. Self-heals once the role exists + they log in again. Consider moving the try/catch *inside* the loop so one bad role_id doesn't block sibling grants — or just confirm 2000008/2000009 exist in prod op_roles.
2. **N+1 inserts (cosmetic):** one INSERT per roster row in a loop. Rosters are tiny (max ~3 roles/person) so it's irrelevant, but a single multi-row INSERT … ON DUPLICATE KEY UPDATE would be marginally cleaner.
3. **Migration is not in an applied-migrations tracker** (repo has no numbered-migration runner visible) — confirm 007 was actually applied to prod, otherwise the table-missing path means the feature is a silent no-op everywhere.

PR #463, opened 2026-07-02
feat(saps): super-admin SAP management — upload, assign, deliver

mbhewitt:
New super-admin page (`/saps`) to manage Setup Access Passes end to end.

## What it does
- **Upload + split** — one batch PDF → per-pass files (`SAP_{date}_{hash}.pdf`), parsing arrival date / year / Ticket ID from the text layer (no OCR). Re-upload dedupes on Ticket ID.
- **People list** — every active volunteer + off-book people, sorted by first-shift arrival, with a "what's missing" to-do column (reuses the volunteer info endpoint's day-requirement logic).
- **Auto assign** — SAP dated the day *before* the first CSP-bearing shift; falls back to the closest earlier day in stock; "Staff" for staff.
- **Deliver + lock** — assign is atomic; download or email flips the pass to `received` and locks it. Re-issuing to someone already served **burns** the old pass. Already-received passes can be re-downloaded / re-emailed.
- **Off-book** — add people by email; they link to a `shiftboard_id` on first Okta login and any email-assigned SAPs move over.
- **Disposition list** — all SAPs for the current burn year (Oct–Sep) with status.

## Schema
`migrations/008_sap_management.sql` (idempotent): extends `op_saps` into a pool, adds `op_sap_offbook`, and adds a generic attachment to `op_email_queue`.

## Notes / judgment calls
- `pdfjs-dist` pinned to **v4** (Docker is Node 22 but local dev is Node 20; v4 installs cleanly on both, same API).
- SAP emails are transactional (an access credential) so they **bypass** the marketing unsubscribe filter.
- `pdfjs-dist`/`pdf-lib`/`formidable` marked `serverExternalPackages` in `next.config`.

## Verification
- 26 unit tests pass (burn-year mode, auto-date pick, 0-CSP exclusion).
- Production build clean.
- Manual e2e against a running dev server with the real 2025 batch: 97 pages → 97 SAPs; re-upload 0 new/97 dup; assign→issue→409-locked; email→queued **with PDF attached**; burn-on-reissue verified; off-book add→assign→unassign.


mbhewitt:
## Independent review (automated) — no critical/high

**Verified sound:** every `/api/saps/*` route is super-admin gated server-side; lock integrity (a `received` SAP can't be reassigned/unassigned; burn-on-reissue only fires for a received current, inside the txn); SQL fully parameterized (only `target.col` is interpolated, constrained to two literals); mail attachment plumbing doesn't regress the ICS path; the `info` endpoint refactor is byte-for-byte equivalent; no path traversal (filenames are server-generated).

**Findings + resolution:**
1. **(medium) Double-assign race** — two concurrent assigns for the *same person* could each reserve a pass (a matching-no-rows `FOR UPDATE` only takes a compatible gap lock). **Fixed** with a per-target `GET_LOCK`/`RELEASE_LOCK` around the txn. Verified: 2 concurrent auto-assigns → 1 active SAP, no leaked locks, pool not wedged.
2. **(low) Off-book Okta link non-atomic** — **Fixed** by moving the SAP-migration UPDATE before the link-flag UPDATE, so a failure retries next login instead of stranding email-assigned SAPs.
3. **(low) Upload orphan files on mid-batch failure** — accepted (re-upload is idempotent on `ticket_id`; orphans are harmless). Marked with a `ponytail:` comment noting the upgrade path.

Also added the requested **resend** path (re-email an already-issued SAP).

PR #464, opened 2026-07-02
fix(shifts): mobile description formatting — stack labels, render **bold** + `---` dividers

mbhewitt:
Per Chipper + Mew 2026-07-01 (mobile testing). **Before/after screenshots approved by Mew.**

### Changes
- **Mobile label overlap** (`ShiftVolunteers.tsx`): the Details/Meal/Notes label column was a fixed `size={2}` (~16% width), so on a phone the label overflowed into the paragraph. Now responsive — `{ xs: 12, sm: 2 }` / `{ xs: 12, sm: 10 }` — so the label stacks above the text on phones, stays side-by-side on desktop.
- **Inline bold** (`FormattedText.tsx`): `**text**` now renders as bold (`<strong>`) instead of showing the literal asterisks.
- **Dividers** (`FormattedText.tsx`): a line of only `---` / `***` / `___` renders as a `<Divider>`, so two notes stored in one field can be visually separated.

### Validation
- `tsc --noEmit` clean.
- Rendered before/after at phone width, signed in — approved by Mew.

Separate follow-up: the mobile **table-overflow** fix (the "outside the black box" DataTable issue).

### DIFF
--- a/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
+++ b/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
@@ -682,12 +682,12 @@ export const ShiftVolunteers = ({
                             <Divider sx={{ my: 2 }} />
                           </Grid>
                         )}
-                        <Grid size={2}>
+                        <Grid size={{ xs: 12, sm: 2 }}>
                           <Typography component="h3" variant="h6">
                             {row.label}
                           </Typography>
                         </Grid>
-                        <Grid size={10}>
+                        <Grid size={{ xs: 12, sm: 10 }}>
                           <FormattedText text={String(row.value)} />
                         </Grid>
                       </Fragment>
--- a/client/src/components/general/FormattedText.tsx
+++ b/client/src/components/general/FormattedText.tsx
@@ -1,16 +1,32 @@
-import { Box, Typography } from "@mui/material";
-import type { ReactNode } from "react";
+import { Box, Divider, Typography } from "@mui/material";
+import { Fragment, type ReactNode } from "react";
 
 // Renders a plain-text description (shift_details / position_details /
-// shift notes) with real paragraph spacing and bullet lists.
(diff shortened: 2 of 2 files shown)


mbhewitt:

Reviewed `git diff origin/main...copy/description-formatting`. Display-only change; no tests demanded.

**FormattedText.tsx**

✅ Divider vs bullet disambiguation is correct. Divider regex `/^(-{3,}|\*{3,}|_{3,})$/` only matches a line that is *entirely* 3+ repeat chars, so `---` (no space) hits the divider branch; `- text` fails it (space + letters) and falls to the bullet regex `/^[-•]\s+(.*)$/`. Divider is also checked *before* the bullet match, so ordering is safe. `--` (2 dashes) matches neither and renders as a literal paragraph — fine.

✅ Bold split is safe across all cases: no bold → single Fragment; one span → `[text, **b**, text]`; multiple spans → interleaved (empty-string parts become harmless empty Fragments); stray unmatched `**` (e.g. `a **b`) → no capture, rendered literally, no throw. `[^*]+` is linear, no ReDoS.

✅ React keys present on every mapped node — `key={index}` on both `<strong>` and `<Fragment>` in `renderInline`, and on bullet `<li>`s. Block-level keys (`index` for paragraphs/dividers, `ul-${index}`/`ul-end` for lists) stay unique across siblings; no key-warning risk.

✅ Bullets still render and now support inline bold via `renderInline(bullet)`.

⚠️ Minor, acceptable: a literal `***` / `___` line renders as a divider rather than text, and bold cannot span a line break (lines are split first). Both are documented/expected for these plain-text fields.

**ShiftVolunteers.tsx**

✅ `size={{ xs: 12, sm: 2 }}` / `size={{ xs: 12, sm: 10 }}` is valid MUI 7 Grid syntax (same file already uses the new `size` API). Only the two label/value columns changed; nothing else touched.

Verdict: ✅ Correct and safe to merge — no bugs found, only cosmetic edge cases that are by design.

Issue #460, opened 2026-07-01
New-volunteer onboarding flow: can't find their way to shift signup


mbhewitt:
From Chipper 2026-07-01, motivated by watching a real volunteer ("Elderbert") try to sign up.

### The saga (motivating example)
- Sent him the shift-signup link; he couldn't log in (didn't know his Burner Profile credentials).
- Days later, updated his Burner Profile, but the session had expired and he'd lost the link.
- Days after that he finally logged in — and **could not find his way to shift signup**. First thing he clicked took him **back to his Burner Profile**, where he started editing again until the **session expired**.
- Got him back in; he read + signed Behavioral Standards, then went to Training — which for most new volunteers is empty/confusing because **training doesn't show until they've signed up for shifts**.
- He never did reach shift signup.

### Problems surfaced
1. **No obvious path to shift signup.** (Being addressed now: a permanent, always-first checklist item "View/Add Census shifts" → `/shifts`.)
2. **Session expires mid-flow**, repeatedly. (Separate change in progress: longer session for off-playa/web; the complete fix is a sliding/renewing session.)
3. **First click bounces back to Burner Profile** instead of moving forward into the app.
4. **Training section shown before any shifts are signed up** — empty/confusing for new volunteers; consider hiding or reframing until relevant.

### Goal
A new volunteer landing in the app should be guided, unmistakably, to sign up for shifts first — regardless of session state or where they land.

Related: session-timeout change (off-playa) and the permanent checklist item (both in progress).

PR #465, opened 2026-07-02
fix(session): 24h session for off-playa web; keep 1h on-playa

mbhewitt:
Volunteers signing up from their own devices off-playa were getting logged out mid-flow by the 1-hour session (the Elderbert saga, #460). This splits session lifetime by deployment:

- **Off-playa** (`NEXT_PUBLIC_PIN_ENABLED="false"`, Okta-only) → **24h**
- **On-playa** lab tablets → **1h** (unchanged — shared devices; a set-down tablet shouldn't leave someone logged in as another volunteer/admin)

`NEXT_PUBLIC_PIN_ENABLED` is inlined at build time, so this is decided per build (prod off-playa build → 24h; on-playa build → 1h). One-line-ish change in `client/src/lib/session.ts`.

**Approved by Mew 2026-07-02.** `tsc --noEmit` clean.

Note: this is an *absolute* session lifetime (not sliding). The fuller fix for "logged out while actively working" is a sliding/renewing session — tracked separately as a future item.

### DIFF
--- a/client/src/lib/session.ts
+++ b/client/src/lib/session.ts
@@ -1,7 +1,17 @@
 import crypto from "crypto";
 
 const COOKIE_NAME = "census-session";
-const SESSION_DURATION_MS = 60 * 60 * 1000; // 1 hour
+// Session lifetime differs by deployment (per Mew 2026-07-02):
+//   - On-playa lab tablets are SHARED — keep the window short so a set-down
+//     tablet doesn't leave someone logged in as another volunteer/admin.
+//   - Off-playa (Okta-only, NEXT_PUBLIC_PIN_ENABLED="false") volunteers sign
+//     up from their own devices at home, so a 1-hour window just logs them
+//     out mid-flow. Give them a full day.
+// NEXT_PUBLIC_* is inlined at build time, so this is a per-build decision.
+const SESSION_DURATION_MS =
+  process.env.NEXT_PUBLIC_PIN_ENABLED === "false"
+    ? 24 * 60 * 60 * 1000 // 24 hours off-playa (web)
+    : 60 * 60 * 1000; // 1 hour on-playa (shared tablets)
 
 interface SessionPayload {
   shiftboardId: number;


mbhewitt:

Reviewed `client/src/lib/session.ts` (`git diff origin/main...fix/session-timeout-offplaya`). Scoped, low-risk change; approve.

✅ **Cookie `Max-Age` and signed `expires` stay consistent.** In `buildSessionCookie`, both `expires` (`Date.now() + SESSION_DURATION_MS`) and `maxAgeSeconds` (`SESSION_DURATION_MS / 1000`) derive from the same constant, so lengthening the window keeps the HMAC payload and the browser cookie lifetime in lockstep. No drift.

✅ **Env gate matches the rest of the app.** Every other on/off-playa check uses `process.env.NEXT_PUBLIC_PIN_ENABLED !== "false"` to mean *on-playa* (`Home.tsx`, `middleware.ts`, `Header.tsx`, `Footer.tsx`, `SignIn.tsx`, `sign-in/index.ts`). This PR's `=== "false"` is the correct inverse for *off-playa* → 24h. `"false"` is the documented off-playa sentinel (unset/"true" = on-playa → 1h). Consistent.

✅ **Env value is reliably present at build time.** `NEXT_PUBLIC_*` is inlined at build. All importers of `session.ts` are server-side (`pages/api/auth/okta/callback.ts`, `pages/api/sign-in/index.ts`, `pages/api/auth/sign-out.ts`, `lib/withAuth.ts` via `NextApiRequest`), and the module imports node `crypto`, so it can only run server-side — the inlined value is present in that bundle. Since the flag is baked per build (playa vs off-playa are separate builds), the timeout is correctly a per-build decision.

✅ **No other behavior changed.** `sign`/`verify`/expiry logic untouched; `verify` still rejects `expires < Date.now()` and constant-time-compares the HMAC. 24h on a personal device for a volunteer signup flow is reasonable and only applies off-playa; shared on-playa tablets keep the 1h window.

Verdict: ✅ Approve — consistent, correctly gated, server-side-safe; no security regression.

PR #471, opened 2026-07-03
fix(session): off-playa web session 12h (was 24h)

mbhewitt:
Per Mew 2026-07-03. Drops the off-playa/web session from 24h to **12h** — still a generous buffer for a home signup session, without holding sessions open a full day. On-playa lab tablets unchanged (1h). One-value tweak to #465 (already reviewed).


mbhewitt:
Closing — Mew confirmed 24h session is the keeper (2026-07-07). The 24h off-playa session + 24h idle timeout are already live; no need for the 12h change.

Issue #470, opened 2026-07-02
Shift signup: let volunteers see their already-signed-up shifts (calendar/conflict view) while adding shifts


mbhewitt:
Volunteer feedback via Chipper 2026-07-02 (from Heartspace, app-feedback Discord channel):

> "I have used the system both on my phone and computer to sign up for various shifts. My main feedback is that when I go to add a new shift it's not obvious to me where I can see what I have already signed up for. I was able to do that in Shiftboard. It seems that I need to go back out to my account to see all the shifts I've signed up for."

### The gap
When a volunteer is adding shifts, there's no in-context view of what they've **already** signed up for. They have to leave the signup flow and go back to their account. Shiftboard showed this as a **calendar**: your signed-up shifts in **green**, and shifts that conflict (same day/time block) in **red**, so it's obvious you shouldn't double-book. (Chipper has a Shiftboard screenshot to attach.)

### Why existing/planned pieces don't fully cover it
- The **double-booking block** (#424) *prevents* booking a conflicting shift at submit, but doesn't *show* the volunteer their existing shifts or the conflict beforehand.
- The planned **inline-shifts-on-checklist** idea (#461 restructure) only helps volunteers **with requirements** — many volunteers have none (they just want to volunteer, no camp/SAP/ticket), so they'd have no inline shifts. And it's not a calendar.

### Options to consider
- Smaller win: surface the volunteer's current shifts (and same-day conflicts) directly in the add-shift flow.
- Bigger: a Shiftboard-style **calendar view** (signed-up = green, conflicting = red) during signup.

Likely next-season given scope, but capturing the real user need now. Related: #424 (double-booking), #452 (gray out conflicting shifts), #461.

mbhewitt:
Chipper's refined design direction (2026-07-02) — "yes, and":

- **Option A (Chipper's lean):** surface **available shifts inline within the checklist sections** on the account/info page (ties into the #461 restructure). Caveat already noted: does nothing for volunteers with no requirements (no inline shifts).
- **Option B (the "yes, and"):** also offer a **"View your calendar"** that shows the volunteer's **assigned shifts**, with a toggle to **also show available shifts** — ideally using the **same filter criteria as the Shifts List page**. Green = yours, and available ones selectable (Shiftboard-style).

Both can coexist: inline-in-checklist for the guided path, plus an opt-in calendar for the "show me everything and let me pick around what I have" path.

[...]

Issue #458, opened 2026-07-01
Role-restricted shifts: Add-volunteer dialog opens for ineligible users; gate is client-side only


mbhewitt:
Reported by Chipper 2026-07-01. **Low priority / future — no immediate change requested.**

### What Chipper saw (UX inconsistency)
On a role-restricted shift (e.g. **Census Lab Stewardship** → the *Census Lab Steward* position, meant only for volunteers camped at Census Lab / holding the `CensusLabCamp` role), the **"Add volunteer" button is only disabled when the shift is full.** A role-ineligible volunteer can still *open* the Add-volunteer dialog — they just can't complete the add, because every position is grayed out (disabled `MenuItem`). So it's a dead-end dialog rather than a blocked button, unlike full shifts which block entry entirely.

**Normal use is safe:** a regular user clicking through the UI genuinely cannot sign themselves up for a position they're not eligible for. The position gray-out + required-field validation prevent it.

### The deeper reason (defense-in-depth gap)
The role-eligibility gate is enforced **only in the browser**:
- `ShiftVolunteersDialogAdd.tsx` disables the position `MenuItem` when `roleRequiredId !== 0` and the volunteer lacks that role.
- The server POST handler `client/src/pages/api/shifts/[timeId]/volunteers/index.ts` does **not** check `roleRequiredId`, nor that the requester is the volunteer themselves or an admin. It only requires auth (`withAuth`) and blocks canceled shifts.

So a crafted/stale-tab `POST` could assign a role-required position to any `shiftboard_id`. Low likelihood (needs a hand-built request, not reachable by normal clicking), but it means the "only campers can take stewardship" rule isn't actually *guaranteed* — it's a UI convention.

### Suggested fix (when this is picked up)
1. **UX consistency:** at the shift list/detail level, treat "no position available to this volunteer" the same as "shift full" so the Add entry is disabled/hidden consistently.
2. **Server enforcement:** in the POST handler, verify the target volunteer holds `roleRequiredId` for the position (and that the requester is self-or-admin) before insert — mirroring the existing server-side canceled-shift check.

Related: #445 (block ineligible signups), #452 (gray-out vs hide), #429 (Camp Steward checklist).

mbhewitt:
[...]

PR #475, opened 2026-07-04
feat(schedule): My Shifts / shift-view — agenda with open shifts, conflicts, eligibility (#470)

mbhewitt:
**The shift-view feature (#470)** — greenlit for this season, built autonomously in 4 vertical slices. New route `/volunteers/[shiftboardId]/schedule`: a day-grouped **agenda** (approved mockup), phone + desktop.

### Slices
1. **My Shifts (read-only)** — the volunteer's signed-up shifts, day-grouped; green cards, canceled state, empty state. Reuses `/api/volunteers/[id]/shifts`.
2. **Open shifts + toggle + filter + sign-up** — merges `/api/shifts`; "Show open shifts" toggle, department filter, legend; open cards show slot counts + "Sign up" (links to shift detail — reuses existing add flow + double-booking guard); dedups shifts already held; "Full" state.
3. **Conflict flagging** — open shifts overlapping one you're on gray out with "⚠ Same time as your <X> shift"; client-side strict overlap. Catches cross-department double-booking Shiftboard doesn't.
4. **Role-eligibility** — NEW read-only endpoint `/api/volunteers/[id]/shift-eligibility` (self-or-admin gated) returns role-gated shifts; those gray out with "🔒 Requires: <role>" + "Not eligible", no sign-up. **Does not touch `/api/shifts` or the signup path.**

### Safety / scope
- **No shared code modified** except adding the new page + new endpoint. `/api/shifts`, `getShiftList`, signup POST all untouched.
- Eligibility is **non-blocking** (degrades to "all eligible" if it doesn't load).
- `tsc` clean; **production `next build` passes**; authed previews per slice; independent reviews per slice + a comprehensive final review.

### Deferred (need Mew)
- **Server-side signup enforcement** (hard block on ineligible POST) — stays in **#458**; this PR is the display layer.
- **Nav link** to reach /schedule — placement is Chipper's product call; route works via direct URL meanwhile.
- Friendly exclusion-reason **config** on roles (#470/#472) — uses the auto role-name fallback for now.

Authority: Slices 1–3 UI (Chipper); Slice 4 touches eligibility (Mew's deploy nod). Staged, not deployed.

### DIFF
--- a/client/src/app/volunteers/[shiftboardId]/schedule/Schedule.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/schedule/Schedule.tsx
@@ -362,6 +362,12 @@ export const Schedule = ({ shiftboardId }: IScheduleProps) => {
                       : item.state === "open"
                         ? theme.palette.secondary.main
                         : theme.palette.divider;
+                  // <=10% filled = genuinely needs help (matches the reports
+                  // sheet's UNDER=0.10 threshold, per #470).
+                  const underfilled =
+                    !!item.slots &&
+                    item.slots.total > 0 &&
+                    item.slots.filled / item.slots.total <= 0.1;
                   return (
                     <Card
                       key={item.key}
@@ -499,7 +505,9 @@ export const Schedule = ({ shiftboardId }: IScheduleProps) => {
                                 <Box
                                   component="span"
                                   sx={{
-                                    color: theme.palette.secondary.main,
+                                    color: underfilled
+                                      ? theme.palette.error.main
+                                      : theme.palette.secondary.main,
                                     fontWeight: 800,
                                   }}
                                 >
@@ -507,7 +515,7 @@ export const Schedule = ({ shiftboardId }: IScheduleProps) => {
                                     ? item.slots.total - item.slots.filled
                                     : 0}
                                 </Box>{" "}
-                                open
+                                open{underfilled ? " · needs help" : ""}
                               </Typography>
                               <Button
                                 component={Link}
--- a/client/src/app/volunteers/[shiftboardId]/schedule/Schedule.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/schedule/Schedule.tsx
@@ -8,17 +8,25 @@ import {
   CardContent,
   Chip,
   Container,
+  FormControl,
+  FormControlLabel,
+  InputLabel,
+  MenuItem,
+  Select,
   Stack,
+  Switch,
   Typography,
 } from "@mui/material";
 import { useTheme } from "@mui/material/styles";
 import Link from "next/link";
+import { useMemo, useState } from "react";
 import useSWR from "swr";
 
 import { BreadcrumbsNav } from "@/components/general/BreadcrumbsNav";
 import { ErrorAlert } from "@/components/general/ErrorAlert";
 import { Loading } from "@/components/general/Loading";
 import { Hero } from "@/components/layout/Hero";
+import type { IResShiftRowItem } from "@/components/types/shifts";
 import type { IResVolunteerShiftItem } from "@/components/types/volunteers";
 import { fetcherGet } from "@/utils/fetcher";
 import { formatDateName, formatTime } from "@/utils/formatDateTime";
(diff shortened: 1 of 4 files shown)


mbhewitt:

Reviewed `git diff origin/main...feat/my-shifts-agenda` — new read-only agenda route (`schedule/page.tsx` + `Schedule.tsx`). Low blast radius, no real bugs found.

**✅ Correctness**
- All fields used exist on `IResVolunteerShiftItem`: `shift.{date, dateName, startTime, endTime, positionName, csp, canceled, timePositionId}` and `department.name`. Confirmed against `components/types/volunteers/index.ts` and the API mapper in `pages/api/volunteers/[shiftboardId]/shifts/index.ts`.
- `groupByDay` is correct for the date-ordered API response (`ORDER BY date, start_time_text`). Empty `[]` → returns `[]` → empty state; `!data` guard runs before grouping, so no null/crash path.
- Loading / error / empty states all present and reasonable (`Loading`, `ErrorAlert`, empty-state CTA card).

**✅ React**
- Card key `shift.timePositionId` is the `op_volunteer_shifts` signup PK — unique per row; a volunteer cannot hold the same `time_position_id` twice. Day key `day.date` is unique because consecutive same-date rows are merged into one bucket. No key-collision warnings.

**✅ Consistency & auth**
- `page.tsx` mirrors `info/page.tsx` exactly (AuthGate + `ACCOUNT_TYPE_AUTHENTICATED` + `await params`).
- Self-view is safe on both layers: server `isOwnerOrAdmin` allows `session.shiftboardId === requested` (own record) or admin; client `AuthGate` requires an authenticated session. No IDOR.
- Imports all resolve, banner asset `public/banners/camp-at-day.jpg` exists, no route collision (only `schedule/` dir). The `account/VolunteerShifts.tsx` table view consumes the same endpoint — this is a lighter parallel presentation, not harmful duplication.

**⚠️ Nits (non-blocking)**
- `groupByDay(data)` groups the full list incl. canceled, but the empty-state and footer count use `active` (non-canceled). A volunteer whose *only* shifts are canceled sees a page of CANCELED cards, no footer count, and the empty-state CTA never shows — arguably intended, but the "Everything you're signed up for" header reads slightly off in that case.
- The `useSWR` destructure is typed `data: IResVolunteerShiftItem[]` (non-optional) though SWR returns `undefined` until loaded — the `!data` guard makes this safe at runtime; type is just slightly optimistic.
- If a signup ever has no matching `op_dates` row (LEFT JOIN → null date), `formatDateName` renders "Invalid Date" rather than crashing. Very low risk with real data.

**Verdict: ✅ Approve — read-only UI, correct, safe, no changes required (nits optional).**

mbhewitt:
🤖 Comprehensive review (automated)

Reviewed `feat/my-shifts-agenda` vs `origin/main` — 3 new files, 670 insertions, **zero deletions, no shared code modified** (verified via `git diff --numstat`). All imports/types/utils resolve.

### 1. Eligibility endpoint — SQL & security ✅
`shift-eligibility/index.ts` is correct.
- **Authz:** `withAuth` (session required) + `isOwnerOrAdmin(session, Number(shiftboardId))` → owner-or-admin object-level gate. No IDOR; a non-owner gets 403. Read-only (GET only; everything else 404).
- **Injection:** the only interpolated value (`shiftboardId`) is bound as a `?` param cast to `Number`. Safe.
- **Takeable logic matches spec:** `CASE WHEN pt.role_id IS NULL OR pt.role_id = 0 THEN 1 WHEN vr.shiftboard_id IS NOT NULL THEN 1 ELSE 0`, with the `op_volunteer_roles` LEFT JOIN correctly filtered `add_role = true AND remove_role = false`. Per-shift aggregation ORs `takeable` across positions and only emits a shift when **no** position is takeable — exactly "every position needs a role you lack." Column/flag names (`remove_shift_time`, `canceled`, `remove_time_position`, `delete_position`) all match usage elsewhere in the codebase.
- **Perf:** single query, no N+1. Duplicate `op_volunteer_roles` rows can't corrupt the result (OR-aggregated). Shifts with zero surviving positions fall through to "takeable" (safe default — never over-restricts).
- Note this is *stricter* than `isAdmin()` (which checks only `remove_role=false`), but stricter is the safe direction here.

### 2. Client merge logic ✅
- **Dedup:** `myTimeIds` skips open rows already held. `/api/shifts` aggregates one row per `shift_times_id`, so a held multi-position shift shows once as "mine" and its remaining open slots are intentionally suppressed — reasonable product call.
- **Conflict:** `overlaps()` is strict (`isBefore` both ways) so back-to-back shifts don't clash. Both sides use the same `start_time ?? start_time_text` format, so the dayjs comparison is apples-to-apples.
- **State priority:** ineligible > conflict > full > open, exactly as specified.
- **Keys unique:** `mine-${timePositionId}` / `open-${o.id}` — disjoint namespaces.
- **Non-blocking eligibility:** `dataElig?.[o.id]` + `Array.isArray` guard; undefined/errored elig can't crash, just doesn't gray anything. Loading gate waits on mine+open only (intended).
- **useMemo deps:** `[dataMine, dataOpen, dataElig]` and `[items, showOpen, department]` both complete.

### 3. Regression risk ✅
Purely additive: new route `/volunteers/[shiftboardId]/schedule` + new endpoint. No existing file touched. Page uses `AuthGate` (client, authenticated-only) but the real protection is the server-side `isOwnerOrAdmin` on both data endpoints — correct defense-in-depth, no PII leak to non-owners.

### ⚠️ Minor (non-blocking, no fix required to ship)
- **No server-side signup enforcement:** an ineligible/conflicting volunteer is only *visually* gated; direct navigation to `/shifts/{id}/volunteers` still allows signup. Explicitly deferred to #458 in the endpoint comment — fine for a read-only display PR, just don't let #458 slip.
- **Brief flash:** shifts render before `dataElig` resolves, so an ineligible shift can momentarily show "Sign up" then flip to "Not eligible." Acceptable given non-blocking design.
- **`/api/shifts` returns Training + canceled shifts**; the page relies on the client filter (`o.canceled` skip) but Training-department shifts will surface as "open" here. Pre-existing endpoint behavior, not this PR — flag only if "My Shifts" shouldn't list trainings.

### Verdict: ✅ SHIP
Backend authz and SQL are correct and injection-safe; client merge/state logic is sound; zero blast radius on existing pages. Minor items above are tracked/acceptable, none blocking.

Issue #478, opened 2026-07-05
Mobile production bugs (prod /info + /shifts on phones)


mbhewitt:
From Chipper 2026-07-05, testing prod on an Android phone. Grouped so we can check them off.

1. **🔴 SHOWSTOPPER — trapped ⋮ menu + back-button escape.** The Assigned-Shifts table (`VolunteerShifts`, a `DataTable`) overflows the screen on mobile; the "actions" ⋮ column is off-screen. Tapping ⋮ opens the menu off-screen and scroll-jumps up to the checklist — the user is stuck. Their only escape is the back button, which (after Okta login) walks them **out of the app to the Burner Profile login page** (Okta redirect history). Two compounding bugs: table overflow (root of the trap) + OAuth back-escape.
2. **Passcode reveal — logo stretched to an ellipse.** `PasscodeReveal` renders the CENSUS logo stretched (should stay circular) on mobile.
3. **Filters popover cut off.** Hardcoded `minWidth: 400px` on `MUIDataTableFilter-root` (in the shared `DataTable` theme) is wider than a phone, so filter options run off-screen on `/shifts`.
4. **"Get involved" sidebar at the bottom on mobile.** The optional-links right sidebar on `/info` stacks to the very bottom on phones (Grid xs:12) — should be findable near the checklist, not after scrolling the whole page.

Fix approach: mobile card-stacking for the shared `DataTable` fixes #1's overflow (and the off-screen ⋮) + relates to #459; #3 is a quick responsive width fix; #2 an aspect-ratio fix; #4 a mobile layout/order fix; the OAuth back-escape (#1b) needs post-login history cleanup.

mbhewitt:
### OAuth "back-button escapes to Burner Profile" — analysis + options for Mew

**Root of the escape:** the `login.burningman.org` (Burner Profile / Okta) page is a real entry in the phone's browser history. Pressing "back" from the app lands there, and that page has no idea it should return to our app (the user arrived by hitting back, bypassing the return-to handoff our `/api/auth/okta` normally sends).

**What's already good:** our post-login landing (`AuthComplete`) uses `router.replace`, so `/auth/complete` isn't left in history. The residual entry is BM's login page, which is **cross-origin — we can't remove it or change what it does** once the user is on it.

**The real fix (this PR):** eliminate the ⋮ trap so users aren't *forced* to hit back. That removes the most common trigger. But (per Chipper) people tap back arbitrarily, so it's not a complete guarantee.

[...]

Issue #459, opened 2026-07-01
Shift list: make column meaning clear on mobile (CSP label/info, sticky headers, de-emphasize CSP)


mbhewitt:
From Chipper 2026-07-01 (mobile testing). **Design call for Captain Mew; largely next-year.**

On a phone, the shift list table is wider than the screen, so you scroll horizontally and **lose the column headers** — leaving an unlabeled number/range (CSP) prominently in view that nobody recognizes (some assume it's signups).

### Near-term (can do now, small)
- **De-emphasize CSP / prioritize signups:** flip the **CSP** and **Filled / Total** columns so the signup count sits prominently (right after Time) and CSP moves to the right. Simple column reorder in `Shifts.tsx` (verified safe — only `rowData[1]` is referenced elsewhere). *Chipper flagged this as a Mew design decision since it changes a screen Mew laid out.*

### Next-year wishes (Mew to decide)
- **Explain CSP in the header** — spell it out and/or add a small info button/tooltip ("Census Shift Points") so people know what the number means.
- **Sticky column headers** — keep headers visible while scrolling the list so users don't have to scroll back to the top to know what they're looking at. (Ties into the broader mobile-responsiveness fix for the shared `DataTable`.)

Related: the shared `DataTable` (mui-datatables, `responsive: "standard"` + `resizableColumns`) overflows the viewport on mobile — separate fix in progress.

mbhewitt:
Additional rationale from Chipper (2026-07-01) for whoever (Mew) decides the CSP treatment:

- Mew specifically wanted CSP visible on this screen; Chipper originally preferred not to show it until a volunteer clicks into a shift's details.
- Chipper's core objection: CSP sits "right smack in the middle of columns that actually tell them things about signing up," with no visible header on mobile — so it reads as a random number/range and gets mistaken for signups.
- Bigger-picture concern Chipper voiced: CSP/perks shouldn't be the primary *driver* for volunteering. The ideal is people sign up for the mission and are then pleasantly surprised the perks are good — not sign up chasing points. Worth weighing when deciding how prominently to display CSP.

This is context only; the flip + display decision remains Mew's call.

mbhewitt:
Mobile table-overflow exploration (2026-07-02, Chipper feedback):

[...]

PR #479, opened 2026-07-05
fix(mobile): showstopper trap + passcode logo + filters popover + sidebar links (#478)

mbhewitt:
Prod mobile bugs from Chipper 2026-07-05 (#478):

- **🔴 Showstopper trap:** shared `DataTable` now **stacks into labeled cards on phones** (`responsive: vertical`, `resizableColumns: false`). The Assigned-Shifts table's ⋮ 'actions' was hidden off-screen and trapped users; now every field incl. the ⋮ is reachable per card. Desktop unchanged. (Also the #459 answer.)
- **Filters popover:** `minWidth: 400px` -> `min(400px, calc(100vw - 32px))` so it fits a phone.
- **Passcode logo:** `height: auto` + `aspectRatio: 1/1` so the round logo isn't squashed into an ellipse.
- **Get-involved sidebar:** links reordered (Hive top, Discord bottom); dropped 'Optional:' from the title.

`tsc` clean. Authed mobile previews captured.

Still open in #478 (not in this PR): sidebar mobile *placement*, and the OAuth back-escape (needs Mew — writing up options).


mbhewitt:

Reviewed `git diff origin/main...fix/mobile-bugs`. The mobile fixes are sound; one stray file to drop.

**DataTable.tsx (the shared, high-blast-radius change)**
- ✅ `responsive: "vertical"` is valid — installed `mui-datatables@4.3.0`, and `@types/mui-datatables` declares `Responsive = "vertical" | "standard" | "simple"`.
- ✅ `useMediaQuery("(max-width:600px)")` is SSR-safe. DataTable is only rendered inside `"use client"` consumers (Volunteers, Shifts, Roles, Saps, Dates, ShiftTypes, positions, categories, per-shift/role volunteer tables). MUI returns `false` on the server and first client render, then updates via effect — no hydration mismatch, layout-only. Desktop path (`isMobile===false`) yields `responsive:"standard"`, identical to before.
- ✅ `minWidth: min(400px, calc(100vw - 32px))` is valid CSS and does what it says: on viewports narrower than ~432px the filter popover shrinks to `100vw - 32px` instead of overflowing; desktop stays 400px.
- ⚠️ `resizableColumns: false` **is a desktop behavior change**, not just mobile. Grep shows no consumer references `resizableColumns` (only DataTable sets it), so nothing breaks — but desktop users lose column drag-resize. Intentional per the comment (it fights the responsive layout); calling it out so it's a conscious trade, not a surprise.

**PasscodeReveal.tsx**
- ✅ No distortion. Canvas intrinsic resolution is set in JS (`canvas.width/height = img.naturalWidth/naturalHeight`); CSS only scales display. `width:260 + maxWidth:100% + aspectRatio:1/1 + height:auto` keeps it square and actually fixes the old bug (fixed `height:260` squashed into an ellipse when width shrank below 260 on narrow phones).

**GetInvolved.tsx**
- ✅ Pure data reorder + copy. `key={item.href}` — hrefs are unique, no logic depends on array order.

**❌ Stray file: `client/shot_info.cjs`**
- A local Playwright screenshot/debug script (hardcoded shiftboardId 26302, mints a `census-session` cookie from `.env.local`, points at `localhost:3020`) is committed in this PR and is not gitignored. Harmless (local-only, reads your own env) but doesn't belong in the branch. Please remove it before merge.

**Verdict:** ✅ Approve once `shot_info.cjs` is dropped. The mobile fixes are correct and low-risk across all DataTable pages; only note is the (intentional) loss of desktop column-resize.

Issue #461, opened 2026-07-01
Rethink the big On-Playa arrival/camping fields at top of /info (rely on checklist; re-prompt near SAP)


mbhewitt:
From Chipper 2026-07-01. **Design call for Captain Mew.**

The `/info` (VolunteerInfo) page currently shows a large "On-Playa Information and Early Entry / SAP" section at the **top** with arrival date, early-entry status, and camping-location fields. There's also a "provide your plans" **checklist item** ("plans") that just points up at that section.

### Chipper's proposal
- **Drop the big top fields** and instead collect arrival/camping info **through the checklist item** itself.
- Benefits:
  - More **control over when** volunteers provide it (as a guided checklist step, not a wall of fields up top).
  - Ability to **re-prompt / pop it back up for a recheck** as we get close to SAP send time, so the arrival date is fresh when it matters most (ties into #456 / arrival-date accuracy).

### Related onboarding trap (from #460 / Elderbert)
- There's a link at the very top, next to the grayed-out info, telling volunteers they can update this on their Burner Profile. That's **where Elderbert first clicked and bounced back to his Burner Profile** (and again from the checklist's Burner-Profile item).
- Interim option if we keep the top section for now: **make that top "update on Burner Profile" reference plain text, not a link**, so it stops pulling people off-task. If we remove the top section entirely (above), the link goes away with it.

Scope: this reshapes a page Mew designed, so it's their call. Captures the idea so it isn't lost.

mbhewitt:
Concrete proposal from Chipper 2026-07-02 (the /info page restructure):

**Core goal:** make the page unmistakably read as a **checklist** so volunteers (especially confused ones like "Elderbert", #460) actually work through the steps. "If people realize they're looking at a checklist, they'll use it like one."

1. **Checklist becomes the topmost / most prominent element**, right under the user's basic info. Today it's hidden below the arrival/camping fields, so volunteers don't realize a checklist exists until they scroll.
2. **Move the arrival-date + camping-location fields out of the top section and into their existing "Tell us your plans" checklist item.** Info is still collected (fields expanded by default), just as a checklist step — no more bouncing volunteers back to the top when they reach that item.
[...]

PR #480, opened 2026-07-05
feat(info): checklist to the top; fold arrival/camp into the plans item (#461)

mbhewitt:
Per Chipper + Mew (#461, approved 2026-07-05). Makes the account page unmistakably a checklist.

- Removes the standalone 'On-Playa Information / SAP' top section; the **checklist is now the topmost card**.
- Arrival-date / early-entry / camping fields move **inline into the 'Tell us your plans' checklist item** (expanded, still save on change).
- Drops the unused onPlayaRef; selects go full-width on phones.

Kept the 'plans' item in place this pass — moving it below Training (Chipper's closer-to-playa reasoning) is a trivial follow-up. `tsc` clean; before/after previewed.

### DIFF
--- a/client/src/app/volunteers/[shiftboardId]/info/VolunteerInfo.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/info/VolunteerInfo.tsx
@@ -160,7 +160,6 @@ export const VolunteerInfo = ({ shiftboardId }: IVolunteerInfoProps) => {
 
   // refs
   // ------------------------------------------------------------
-  const onPlayaRef = useRef<HTMLDivElement>(null);
   const passcodeRef = useRef<HTMLDivElement>(null);
 
   // state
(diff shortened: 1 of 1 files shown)


mbhewitt:
🤖 Independent pre-deploy review

Reviewed `git diff origin/main...feat/info-checklist-restructure` plus the resulting file state.

✅ `onPlayaRef` fully removed — 0 remaining references (the old `<Card ref={onPlayaRef}>` section is gone). `useRef` import is still needed (`passcodeRef` at line 163), so no unused-import lint break.
✅ All three moved handlers stay wired: `handleArrivalDateChange` (arrival Select), `handleOtherSapToggle` (early-entry Select, `value === "other"`), `handleLocationChange` (camping onBlur). Definitions and call sites both intact.
✅ `plansDone` unchanged (`arrivalDate !== null && location non-empty`) and still drives both the item label and `done`. Early-entry/SAP was never part of `plansDone`, so no regression there.
✅ No dangling JSX: the old 'in the section above' copy is gone, replaced by the inline `onPlayaFields` block rendered inside the checklist item. `isPreOpen`, `hasOtherSap`, `theme`, `Select`, `MenuItem`, `TextField` all still referenced elsewhere — no orphaned imports.
✅ Phone full-width handled via `minWidth: { xs: '100%', sm: 300 }` on both Selects; camping field is already `fullWidth`.

⚠️ Non-blocking: I verified by static inspection (diff + grep), not a full `npm run build`. The evidence is clean, but a build/visual pass on the /info page before prod is the usual gate for a layout change.

SHIP

PR #484, opened 2026-07-06
fix(session): off-playa idle timeout = 24h (stop the 5-min re-logins)

mbhewitt:
The 24h **session** shipped in #465, but a separate **5-minute idle timer** was still logging off-playa users out after a few minutes idle (the constant re-logins Chipper/Mew hit). This sets the off-playa idle timeout to 24h to match the session; **on-playa keeps the short idle logout** (shared tablets). Per Mew 2026-07-06.

One-line behavioral change in Header.tsx. `tsc` clean.

### DIFF
--- a/client/src/components/layout/Header.tsx
+++ b/client/src/components/layout/Header.tsx
@@ -92,9 +92,17 @@ export const Header = () => {
     );
   };
 
+  // On-playa (shared tablets) keeps the short idle auto-logout so a walk-up
+  // doesn't leave someone else signed in. Off-playa (personal devices) the
+  // idle timeout matches the ~24h session, so people aren't kicked out for
+  // stepping away for a few minutes. Per Mew 2026-07-06.
+  const idleTimeoutMs =
+    process.env.NEXT_PUBLIC_PIN_ENABLED !== "false"
+      ? IDLE_MINUTES * 60 * 1000
+      : 24 * 60 * 60 * 1000;
   useIdleTimer({
     onIdle: handleSignOut,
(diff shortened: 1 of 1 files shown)


mbhewitt:
🤖 Independent pre-deploy review

Reviewed `git diff origin/main...fix/off-playa-idle`.

✅ Env check is correct and consistent: off-playa = `NEXT_PUBLIC_PIN_ENABLED === "false"` → 24h timeout; on-playa (anything else, incl. unset) → short `IDLE_MINUTES`. This matches the existing `isOnPlaya = process.env.NEXT_PUBLIC_PIN_ENABLED !== "false"` on line 121 of the same file, so the convention lines up.
✅ On-playa auto-logout is NOT weakened — when PIN is enabled (the shared-tablet case) the ternary still yields `IDLE_MINUTES*60*1000`, identical to before.
✅ `isDisableIdleEnabled ? undefined : idleTimeoutMs` still works — the disable-idle override is preserved (`undefined` disables the timer), only the non-disabled branch value changed.
✅ 24h = 86,400,000 ms is well under the setInterval/setTimeout 32-bit ceiling (~2.147e9 ms ≈ 24.8 days) react-idle-timer uses internally, so no overflow/instant-fire risk.

⚠️ Non-blocking nit: `idleTimeoutMs` recomputes the same `PIN_ENABLED !== "false"` test that `isOnPlaya` already does 20 lines lower. Could hoist `isOnPlaya` above and reuse it, but it reads fine as-is.

SHIP

PR #524, opened 2026-07-10
fix(fetcher): throw on non-2xx so API errors show ErrorPage, not a white-screen

mbhewitt:
## Problem

Mew hit `Application error: a client-side exception has occurred` loading
`/volunteers/820897/info` in prod. 820897's data is fine (exists, not deleted,
5 roles, clean shifts) and the page renders fine now — the failure was **transient**.

Root cause: `fetcherGet` did `fetch(url).then(res => res.json())` for **any**
status. A non-200 body — a transient 500 (very likely the recurring mysql2 pool
wedge, #prod-pool) or a 404 for a removed volunteer — was handed back to SWR as
`data`, then destructured as if it were a real payload (`roles.includes(...)` →
`undefined.includes` → white-screen), instead of hitting the intended
`if (error) return <ErrorPage/>` path.

Reproduced deterministically on `/volunteers/26302/{account,info}` (26302 is a
permanent 404 — no `op_volunteers` row) and via a simulated 500 on
`/volunteers/820897/info` (the transient case). Both white-screened before.

## Fix

`fetcherGet` now throws on `!res.ok`, so SWR sets `error` **and retries**
transient failures. One shared util, all 34 callers covered.

- `fetcherTrigger` is intentionally left unchanged — `SignIn.tsx` relies on it
  returning the 404 body to inspect `statusCode === 404`.
- Healthy 200s are byte-identical to before (`return res.json()`), so no
  behavior change on working pages.

## Testing (local off-playa prod build, matching prod's `NEXT_PUBLIC_PIN_ENABLED="false"`)

| Scenario | Before | After |
|---|---|---|
| `/volunteers/26302/account` (404) | white-screen | graceful ErrorPage |
| `/volunteers/820897/info` + simulated 500 | white-screen | graceful ErrorPage (SWR retries) |
| healthy pages (`/info`, `/volunteers`) | ok | ok (no regression) |

### DIFF
--- a/client/src/utils/fetcher.ts
+++ b/client/src/utils/fetcher.ts
@@ -1,5 +1,16 @@
-export const fetcherGet = async (url: string) =>
-  fetch(url).then((res) => res.json());
+export const fetcherGet = async (url: string) => {
+  const res = await fetch(url);
+  // Throw on non-2xx so SWR sets `error` (and retries) instead of handing the
+  // error body back as `data`. Without this, a transient API failure (e.g. a
+  // wedged DB pool → 500, or a 404 for a removed volunteer) got destructured
+  // as if it were a real payload and white-screened the page. Callers already
+  // guard with `if (error) return <ErrorPage/>`.
+  if (!res.ok) {
+    const body = await res.json().catch(() => null);
+    throw new Error(body?.message ?? `Request failed (${res.status})`);
+  }
+  return res.json();
+};
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const fetcherTrigger = async (url: string, { arg }: { arg: any }) =>


mbhewitt:
**Independent review (automated):** LGTM — safe to merge.

**1. fetcherGet correctness** ✅ Throws on `!res.ok`, returns `res.json()` on ok. Non-JSON error body handled: `res.json().catch(() => null)` swallows the secondary parse error, falls back to `Request failed (${res.status})`. No unhandled throw.

**2. Blast radius** ✅ Swept all ~32 `useSWR(..., fetcherGet)` callers under `client/src`. Every one guards `if (error) return <ErrorPage/>` (or `<Loading/>` for `!data`). One caller has no error guard — `app/saps/Saps.tsx:73` `poolData` — but it's consumed defensively (`(poolData?.saps ?? []).map`, line 447), so an error path renders an empty list, not a crash. No caller degrades.

**3. Callers reading an error body off fetcherGet** ⚠️ One: `app/training/confirmation/[code]/TrainingConfirmation.tsx` reads `data.statusCode` (lines 77, 129-130) — it explicitly relied on the old behavior of getting the `{statusCode,message}` 404/400 body as `data`. **Not broken by this fix:** on a 404 the fetcher now throws → `error` set → line 124 `if (error) return <ErrorPage/>` fires first. Same visual outcome (ErrorPage). The `.statusCode` branches (77, 128-133) just become dead code. Worth a cleanup follow-up, not a blocker. (SignIn:125 reads `.statusCode` off a `fetcherTrigger` result — untouched, correct.)

**4. SWR retry** Throwing now enables SWR's default `errorRetry` (was silent before). No global `SWRConfig` override found, so exponential backoff up to 5 retries kicks in on 5xx — desirable for a wedged-pool 500, harmless for 404. No stale-data concern.

Verdict: clean fix, root-caused in the shared fetcher. Optional: drop the now-dead `.statusCode` guards in TrainingConfirmation.

PR #525, opened 2026-07-10
feat(shift-detail): hide check-in + admin review before the shift; rename add button for non-admins

mbhewitt:
Two shift-detail tweaks per Mew (2026-07-09).

## 1. Hide check-in + admin review before the shift
Before a shift's check-in window opens (`getCheckInType === SHIFT_FUTURE` →
`isCheckInAvailable === false`), the check-in toggle and the admin review (chat)
icon are meaningless — nobody can be checked in yet. They previously rendered
anyway (the switch just disabled). Now both are gated on `isCheckInAvailable`,
reusing the exact logic that governs when anyone can check *themselves* in.

- Applies to both the desktop DataTable cells and the mobile roster cards.
- The admin ⋮ actions menu (view account / remove) is **unchanged** — still
  available to admins pre-shift.

## 2. Rename the add button for non-admins
Non-admins now see **"Add this shift"** (it's self-signup for them); admins keep
**"Add volunteer"**.

## Testing (local off-playa prod build, future-dated Artist shift = SHIFT_FUTURE)

Verified on a real future shift (`/shifts/63/volunteers`) at mobile width:

| View | Check-in switch | Admin review icon | ⋮ menu | Add button |
|---|---|---|---|---|
| Admin, pre-shift | hidden ✓ | hidden ✓ | shown ✓ | "Add volunteer" ✓ |
| Non-admin, pre-shift | hidden ✓ | n/a | n/a | "Add this shift" ✓ |

During/after the check-in window (`isCheckInAvailable === true`) the switch and
review render exactly as before (unchanged JSX, only the wrapping condition
changed). `tsc` clean via `next build`.

### DIFF
--- a/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
+++ b/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
@@ -483,28 +483,33 @@ export const ShiftVolunteers = ({
         playaName,
         worldName,
         positionName,
-        <Switch
-          checked={isCheckedIn === ""}
-          disabled={!isCheckInAvailable}
-          onChange={(event) =>
-            handleCheckInToggle({
-              shift: {
-                positionName,
-                timePositionId,
-              },
-              volunteer: {
-                isCheckedIn: event.target.checked,
-                playaName,
-                shiftboardId,
-                worldName,
-              },
-            })
-          }
-          key={`${shiftboardId}-shift-volunteer`}
-        />,
-        // if volunteer is admin
+        // Check-in only appears once the shift's check-in window is open
+        // (getCheckInType !== SHIFT_FUTURE); before that it's meaningless.
+        isCheckInAvailable ? (
+          <Switch
+            checked={isCheckedIn === ""}
+            onChange={(event) =>
+              handleCheckInToggle({
+                shift: {
+                  positionName,
+                  timePositionId,
+                },
+                volunteer: {
+                  isCheckedIn: event.target.checked,
+                  playaName,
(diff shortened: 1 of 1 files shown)


mbhewitt:

**Cell/column alignment — OK.** The desktop row builder always returns a 6-element positional array (`ShiftVolunteers.tsx:482-581`). The change swaps the Switch cell for a ternary that still yields exactly one element (`<Switch>` or `""`, L488-509), and the review cell stays a single `isAdmin && isCheckInAvailable && (...)` expression (L512). Cell count is invariant across admin/non-admin and pre/during-shift — no off-by-one introduced. (Non-admin still emits 6 cells against 4 columns via the falsy `isAdmin && …` entries, but that's pre-existing; mui-datatables ignores trailing cells.)

**Gating — correct.** Desktop review = `isAdmin && isCheckInAvailable && (...)` (L512); MoreMenu = `isAdmin && (...)`, NOT check-in gated (L539). Mobile matches: review `isAdmin && isCheckInAvailable` (L875 ctx), MoreMenu `isAdmin &&` only (L882), Switch wrapped in `{isCheckInAvailable && (...)}` (L840).

**During/after shift — unchanged.** When `isCheckInAvailable` is true the Switch renders and stays interactive; the old `disabled={!isCheckInAvailable}` prop was correctly dropped since the control is now hidden pre-window rather than disabled.

**Keys/lint — clean.** Desktop Switch keeps its `key`; the `""` fallback needs none. Mobile Switch had no key before and still doesn't (parent card carries it) — no regression.

Verdict: LGTM.

PR #526, opened 2026-07-10
feat(shift-detail): sort lead positions first, then non-lead critical, then by name

mbhewitt:
Per Mew (2026-07-09): on the shift-detail page, both the **Positions list** and
the **volunteer roster** should lead with lead-flagged positions, then non-lead
critical, then everyone else — and then sort by playa name.

## How
Sorted in SQL via a priority tier:
```sql
CASE WHEN pt.`lead`=1 THEN 0 WHEN pt.critical=1 THEN 1 ELSE 2 END
```
- **Positions** break ties by position name.
- **Roster** breaks ties by playa name, then world name.
- `lead` is a MySQL reserved word → backticked.
- Removed the DataTable default `sortOrder` on both tables so the **desktop**
  view honors the API's priority order (the **mobile** cards already render in
  array order). Column headers remain clickable for ad-hoc re-sorting.

No type/shape changes — the client renders each list in the order the API
returns it.

## Testing (local off-playa prod build, shift 5 — 4 positions mixed lead/critical, 14 volunteers)

Positions order: `Gate Sampling Lead` (lead) → `DataBeast Driver` (non-lead critical) → `Gate Sampler`, `Traffic Tamer`.

Roster order (verified at **both** desktop table + mobile cards):
Cali Green, Dan, Rocket (Gate Sampling Lead) → Jeremiah (DataBeast Driver) → Ben, Christina, … (Gate Sampler/Traffic Tamer, alphabetical).

### DIFF
--- a/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
+++ b/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
@@ -416,10 +416,9 @@ export const ShiftVolunteers = ({
     filter: false,
     pagination: false,
     search: false,
-    sortOrder: {
-      direction: "asc" as const,
-      name: "Name",
-    },
+    // No default sortOrder: the API already returns positions in priority
+    // order (lead first, then non-lead critical, then the rest alphabetically).
+    // Column headers stay clickable for ad-hoc re-sorting.
   };
 
   // prepare datatable volunteers
@@ -577,10 +576,9 @@ export const ShiftVolunteers = ({
     }
   );
   const optionListCustomVolunteers = {
-    sortOrder: {
-      direction: "asc" as const,
-      name: "Playa name",
-    },
+    // No default sortOrder: the API already returns the roster in priority
+    // order (lead positions first, then non-lead critical, then the rest by
+    // playa name). Column headers stay clickable for ad-hoc re-sorting.
   };
 
   // render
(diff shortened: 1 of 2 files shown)


mbhewitt:
**Independent review (automated):** LGTM — correct and safe to merge.

**SQL (index.ts:70-76, 102-109):** `lead` is backticked in both queries; inside the JS template literal `\`lead\`` emits a literal backtick, so MySQL sees the reserved word quoted correctly. Both `lead` and `critical` are `tinyint(1) DEFAULT NULL` (current_schema_v2.sql:138-139), and `NULL=1` evaluates to NULL (not true), so NULL rows correctly fall through to tier 2. Tiering is unambiguous: each roster row maps to exactly one position_type. No mis-tiering risk. COLLATE tie-break preserved.

**DataTable (ShiftVolunteers.tsx:415, 578):** Removing `sortOrder` is correct. mui-datatables v4 renders `data` in array order when no `sortOrder` is set, and the Name/Playa name/Position columns don't set `sort:false`, so headers stay clickable. `optionListCustomVolunteers = {}` is valid — `MUIDataTableOptions` fields are all optional and it's spread into `optionListFinal` (DataTable.tsx:97).

**Mobile:** Cards use mui-datatables' `responsive:"vertical"` (DataTable.tsx:85), which renders the same `data` array — so mobile and desktop share the API order. No divergence.

**Types/imports:** Only object properties removed; no unused imports, no type errors.

PR #528, opened 2026-07-12
fix(shifts): await shift-type add writes so new times/positions persist (#515)

### LINKED ISSUE #515: New shift position won't show up properly when added (Chipper)
Chipper 2026-07-09 (higher priority than the SAP items): trying to **add a shift position for a new role** and can't get it to show up properly. Unclear if it's a bug (position added but not rendering) or a non-obvious config step. Also: needs to add **tech shifts for SimCard** (at their request).

To investigate: the create-shift-position flow (op_position_type / op_shift_time_position wiring, role_id gating, whether a new position needs a role/training link to appear). Get Chipper's exact steps to repro. She'll drive this later but the 'not showing up' part is worth tracing for a bug.

mbhewitt:
Fixes #515. Explains the "I added a shift/position, went back, and it wasn't there" reports from **Chipper** (new position for the DataBeast Technician role) and **Rescue** (new training-video shift time).

## Root cause
The shift-type `PATCH` handler and `handleTimeListAdd` fired their DB writes with `forEach(async …)`. `forEach` doesn't await async callbacks, so the handler returned `200` **before** the `INSERT`s committed. The client refetched and the new time/position wasn't there yet → looked like a silent failure. (The `UPDATE` path had already been fixed to `await Promise.all`; the add/remove paths hadn't.)

## Fix
- `handleTimeListAdd`: both `forEach(async)` → sequential `for…of` (awaited; also removes a `generateId` race).
- `PATCH`: `timeListAdd` → `for…of`; `timeListRemove` + the three `timePosition*` loops → `await Promise.all(...map(...))`. Everything completes before the response.
- **Ordering:** brand-new shift times are inserted **last** (after `dbTimePositionList` is read and the position remove-pass runs). If inserted earlier, their freshly-generated positions would be in `dbTimePositionList` but not the request payload, so the remove-pass would immediately delete the just-added position.

## Testing (local prod build, real PATCH round-trip as admin)
| Scenario | Result |
|---|---|
| Add a position to an existing time | ✅ persists, visible on immediate refetch |
| Add a brand-new time **with** a position | ✅ time + position persist, no phantom removal |
| Existing times/positions | ✅ untouched |

Note for Chipper/Rescue: the position types/roles saved fine; only the shift-link write was lost. Once this deploys, re-add the position/time once and it'll stick.

### DIFF
--- a/client/src/pages/api/shifts/types/[typeId]/index.ts
+++ b/client/src/pages/api/shifts/types/[typeId]/index.ts
@@ -373,51 +373,57 @@ const shiftTypeUpdate = async (req: NextApiRequest, res: NextApiResponse) => {
           );
         }
       }
-      timeListAdd.forEach(
-        async ({ endTime, instance, meal, notes, startTime }) => {
-          const [dbTime] = await pool.query<RowDataPacket[]>(
-            `SELECT shift_times_id
-            FROM op_shift_times
-            WHERE end_time=?
-            AND start_time=?`,
-            [endTime, startTime]
-          );
-          const [dbTimeFirst] = dbTime;
-          dbTimeIdExist = dbTimeFirst?.shift_times_id;
-
-          if (dbTimeFirst) {
-            await pool.query<RowDataPacket[]>(
-              `UPDATE op_shift_times
-              SET
-                add_shift_time=true,
-                meal=?,
-                notes=?,
-                remove_shift_time=false,
-                shift_instance=?
-              WHERE shift_times_id=?`,
-              [meal === "None" ? "" : meal, notes, instance, dbTimeIdExist]
-            );
-          } else {
-            // insert new shift time rows
-            handleTimeListAdd({
-              timeList: timeListAdd,
-              typeId: Number(typeId),
-            });
-          }
-        }
-      );
-      timeListRemove.forEach(async ({ shift_times_id }) => {
-        await pool.query<RowDataPacket[]>(
-          `UPDATE op_shift_times
-          SET
-            add_shift_time=false,
-            remove_shift_time=true,
-            shift_instance=?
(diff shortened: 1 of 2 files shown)


mbhewitt:
**Independent review (automated):** LGTM — ships the fix, no new bugs.

1. **No unawaited writes remain.** Both `forEach(async)` blocks in `handleTimeListAdd` (`types/index.ts:27,84`) are now `for…of`; all four write loops in the PATCH handler (`[typeId]/index.ts:413,483,499,547`) are `await Promise.all(...map())`. The one surviving `forEach` (`[typeId]/index.ts:149`) has a synchronous callback (GET handler) — correct.

2. **Reorder is correct.** `timeListAddNew` (`:384`) collects brand-new times; `handleTimeListAdd` now runs at `:565`, after `dbTimePositionList` is read (`:429`) and the remove-pass (`:547`). Since those positions aren't inserted yet, they can't land in `dbTimePositionList` and get flagged for removal. `dbTimeIdExist` (shared `let`, `:228`) still holds the last existing time's id when read at `:506` — same value the old sequential loop produced, so no regression.

3. **Braces balance;** the big `timePositionListAdd` block (`:499–546`) closes cleanly (`})` + `)`). tsc `--noEmit` passing corroborates.

4. `const timeListAddNew: typeof timeListAdd = []` is valid TS and passes the exact element shape `handleTimeListAdd` expects.

5. **Concurrency:** `generateId` (`utils/generateId.ts:14`) is pre-broken — it fires an un-awaited async re-roll and returns the first random number without confirming uniqueness. `Promise.all` doesn't worsen this (collisions were already unguarded), but the collision risk is real and worth a separate ticket. Not blocking this PR.

PR #532, opened 2026-07-13
fix(agenda): filter "Select all" injected a phantom entry ("Full, Full")

mbhewitt:
Reported by Chipper testing **My Shifts**: selecting Full then "Select all" showed the filter as **"Full, Full"** with **no records**, and returning to "All" behaved oddly.

## Root cause
The filter dropdowns (Timeline / Type / Date / Availability) render "Select all" as a **value-less `MenuItem` with an `onClick`** inside a `multiple` MUI `Select`. So clicking it:
1. injects a phantom `""` into the value array (the Availability `renderValue` maps anything not `"open"` → `"Full"`, hence **"Full, Full"**; the phantom state matches nothing → **zero records**), and
2. races the `Select`'s own `onChange` against the `onClick` toggle.

## Fix
Route "Select all" through the `Select`'s `onChange` via a **sentinel value** (`SELECT_ALL`) instead of `onClick`. `onMultiChange` toggles all/none when the sentinel is present, else stores the picked values. `asArray` also drops empty/phantom entries. Applied to **all four** filters.

## Verified (pure-function test of the exact sequence + tsc)
```
pick Open        -> Open
click Select all -> Open, Full   (toggles to both)
click Select all -> All          (toggles back to empty)
pick Full        -> Full
click Select all -> Open, Full   (was "Full, Full" + no records)
phantom blank    -> Full          (dropped, no "Full, Full")
```

### DIFF
--- a/client/src/app/volunteers/[shiftboardId]/schedule/Schedule.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/schedule/Schedule.tsx
@@ -223,18 +223,32 @@ export const Schedule = ({ shiftboardId }: IScheduleProps) => {
     setMyShiftsOnly(false);
   };
 
-  // Keep a multiselect's Select value as a string[] (MUI hands back a comma
-  // string on native change). Shared by every dropdown below.
-  const asArray = (v: string | string[]) =>
-    typeof v === "string" ? v.split(",").filter(Boolean) : v;
+  // Sentinel value for the "Select all" menu item. It's routed through the
+  // Select's onChange (below) rather than a plain onClick — an onClick fights
+  // the Select's own selection handling and injects a phantom "" entry, which
+  // surfaced as a duplicate "Full, Full" and wedged the filter. Never stored.
+  const SELECT_ALL = "__select_all__";
 
-  // "Select all" toggle for a multiselect: if every option is already picked,
-  // clear it; otherwise select them all.
-  const toggleSelectAll = (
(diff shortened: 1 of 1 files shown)


mbhewitt:

Verified the fix in `Schedule.tsx`. Verdict: correct, ship it.

1. **onMultiChange (L239-249):** all cases right. Pick an option → `set(next)` (filtered). Select all when all picked → `[]`; when none/some → `all`. Sentinel intercepted by the `includes` check before any `set`, so it never lands in state.
2. **Sentinel leak:** none. `asArray` (L234-236) `.filter(Boolean)` in both branches, and the sentinel path stores only `all`/`[]`. `renderValue` never sees `__select_all__`.
3. **All four Selects updated consistently** with correct `all` lists — timeline `["future","past"]` (L627), types `typeOptions` (L657), dates `dateOptions.map(d=>d.dateName)` (L687), availability `["open","full"]` (L719).
4. **MUI:** no issue. Checked state is `selected.indexOf(value)!==-1`; sentinel is never in `value`, so the Select-all item never shows checked. Out-of-range warnings fire on *values* not in options — sentinel is never a value, and the MenuItem is present anyway, so no warning.
5. **toggleSelectAll fully removed** — grep shows zero references. tsc clean.

Minor nit (non-blocking): sentinel `__select_all__` collides only if a real `dateName`/type ever equals that literal — effectively impossible.

PR #536, opened 2026-07-15
fix(agenda): keep your own shifts visible under Type/Availability filters; hide Timeline filter off-playa

mbhewitt:
## What & why
Two fixes to the Shifts agenda (`Schedule.tsx`) from Chipper's 2026-07-15 report.

### 1. Signed-up shift dropped from the top list under a Type/Availability filter
Chipper had a **Type** filter active and her signed-up **"Strike Crew"** vanished from the main list — it only showed in the bottom "My Shifts" box.

**Root cause:** a `"mine"` agenda item carries no shift-type (its `type` is a *department-name* fallback, since the signup payload has no shift-type) and its `state` is `"mine"` (never `"open"`/`"full"`). So a **Type** filter can't match it and an **Availability** filter can't match it — and both were being applied to your own shifts, silently dropping them from the top.

**Fix:** Type/Availability now narrow **browse (open) shifts only**. Your own shifts always stay visible in the top list for context. Timeline/Date (real properties of a mine item) still apply.

The bottom "My Shifts" safety-net section now renders only when a **Date** filter is active — the only remaining narrowing that can hide a mine item from the top — so it no longer double-lists your shifts.

### 2. Hide the Timeline (Present/Future) filter off-playa
There are no past shifts until on-playa, and **"Remove all filters"** left its default "future" chip stuck (had to click the ✕ directly). Gated behind `SHOW_TIMELINE_FILTER = false` — flip to `true` to restore when the on-playa work (past shifts + check-in toggle) lands. Filter *state* stays `["future"]` (correct off-playa behavior); only the control + chip are hidden.

## Testing
- `tsc --noEmit` clean
- `next build` ✓ compiled successfully

## Deploy note
Filter *logic* change → needs @mbhewitt deploy sign-off. Layout (hidden filter) is per Chipper's request. Test-first per project policy — a before/after preview can be produced on request (local agenda renders empty due to past-dated test data; needs seeded future data).

### DIFF
--- a/client/src/app/volunteers/[shiftboardId]/schedule/Schedule.tsx
+++ b/client/src/app/volunteers/[shiftboardId]/schedule/Schedule.tsx
@@ -214,6 +214,14 @@ export const Schedule = ({ shiftboardId }: IScheduleProps) => {
   const [availability, setAvailability] = useState<string[]>([]); // "open" | "full"
   const [myShiftsOnly, setMyShiftsOnly] = useState(false);
 
+  // ponytail: Timeline (Present/Future vs Past) filter is hidden until the
+  // on-playa work lands (past shifts + the check-in toggle). There are no past
+  // shifts off-playa, so the control just confused people — "Remove all
+  // filters" left its default "future" chip stuck (Chipper 2026-07-15). State
+  // stays ["future"] (the correct off-playa behavior); we only hide the UI.
+  // Flip to true to bring the filter back.
+  const SHOW_TIMELINE_FILTER = false;
+
   // Reset every filter to its default (#470 design review, item 8).
   const removeAllFilters = () => {
     setTimeline(["future"]);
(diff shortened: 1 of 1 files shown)


mbhewitt:

Reviewed the diff for correctness/logic regressions only. **No blocking issues found** — the two changes are sound.

**Visibility invariant holds.** Traced every filter combination for "mine" items:
- Type/Availability active, no Date filter → mine items are exempt in `mainDays`, so they stay in the **top** list. Bottom section is correctly suppressed (would just double-list). ✅ Every signed-up shift still visible.
- Date filter active → a mine item outside the selected dates drops from the top, but `mineDays` (Timeline-only) + the new `dateFilterActive` gate surfaces it in the bottom safety net. ✅
- No mine item can fall through both — confirmed the old `anyNarrowing` → `dateFilterActive` narrowing is safe *because* Type/Availability no longer drop mine items from the top.

**Diff matches code**: `type: m.department.name` / `state: "mine"` confirmed at construction (lines 134/138), so the rationale (Type/Availability can never legitimately match a mine item) is accurate.

**No lint/dead-code smell**: `timeline` still read by `inTimeline` + hook deps + the (hidden) Select; `setTimeline` still called in `removeAllFilters`. `anyNarrowing` fully removed, no orphan. React keys unchanged. Hook deps for `mainDays`/`mineDays` unchanged and correct.

**Minor nits (non-blocking):**
1. With **My Shifts toggle ON**, the exemption makes the Type and Availability dropdowns silent no-ops (the `i.state !== "mine"` guard is never true once only mine items remain). This is actually an improvement over the prior behavior (which produced an empty "No shifts match" list), but the two dropdowns now do nothing while the toggle is on with no UI hint. Cosmetic.
2. `SHOW_TIMELINE_FILTER = false` is a compile-time-constant condition; some ESLint configs flag `no-constant-binary-expression`/unreachable JSX. Build passes here so it's tolerated, and it matches the stated "flip to re-enable" intent — fine as-is.

PR #538, opened 2026-07-17
Flip /shifts to the agenda; retire separate My Shifts

mbhewitt:
Makes the agenda the single **Shifts** page and drops **My Shifts** as a separate entity (Chipper sign-off 2026-07-17; Mew pre-approved).

## What changed
- `/shifts` now renders the agenda (`<ShiftsAgenda>` → `<Schedule>`), reading the signed-in volunteer's `shiftboardId` from the session.
- Removed the separate **My Shifts** drawer entry — it's now the My-Shifts filter toggle inside the one Shifts page.
- Old `/volunteers/[id]/schedule` **redirects to `/shifts`** so bookmarks/links still land.
- The old `<Shifts>` data-table is kept one import away as trivial revert insurance (repoint `page.tsx`).

## Why it's low-risk
`Schedule` was already built to handle both a signed-out walk-up (`shiftboardId 0` → browse-only) and a signed-in volunteer, and its cards link to the **same** `/shifts/[timeId]/volunteers` page the old table did — so the **on-playa check-in flow is unchanged**. (This is the flip that was reverted once before as #504; the agenda has since been stabilized via #505/#498.)

## Verification (against production builds, rendered authed)
- **Off-playa / volunteer**: `/shifts` renders the agenda — day headers, slot counts, CSP, 🚫 Full, 🔒 role-gating, ⚠️ conflicts, ✓ "You're signed up". Old table gone. Redirect confirmed. Zero console errors.
- **On-playa / walk-up (unauth kiosk)**: `/shifts` stays open (allowlisted), renders browse-only — no personal markers, My-Shifts toggle hidden. Zero console errors. (This path the old auth-gated My Shifts route never exercised.)
- `tsc --noEmit` clean; `next build` clean for both on-playa and off-playa.

### DIFF
--- a/client/src/app/shifts/ShiftsAgenda.tsx
+++ b/client/src/app/shifts/ShiftsAgenda.tsx
@@ -1,0 +1,21 @@
+"use client";
+
+import { useContext } from "react";
+
+import { Schedule } from "@/app/volunteers/[shiftboardId]/schedule/Schedule";
+import { SessionContext } from "@/state/session/context";
+
+// ponytail: the agenda IS the Shifts page now (the "My Shifts" flip). Schedule
+// was already built to handle both a signed-out walk-up (shiftboardId 0 →
+// browse-only) and a signed-in volunteer (personal shifts + eligibility), so
+// /shifts just feeds it the session id. Revert = repoint page.tsx back to the
+// old <Shifts> table, which is intentionally kept one import away.
+export const ShiftsAgenda = () => {
+  const {
+    sessionState: {
+      user: { shiftboardId },
+    },
+  } = useContext(SessionContext);
+
+  return <Schedule shiftboardId={shiftboardId} />;
+};
--- a/client/src/app/shifts/page.tsx
+++ b/client/src/app/shifts/page.tsx
@@ -1,4 +1,4 @@
-import { Shifts } from "@/app/shifts/Shifts";
+import { ShiftsAgenda } from "@/app/shifts/ShiftsAgenda";
 
 export const metadata = {
   title: "Census | Shifts",
@@ -6,7 +6,7 @@ export const metadata = {
 const ShiftsPage = () => {
   // render
   // ------------------------------------------------------------
-  return <Shifts />;
+  return <ShiftsAgenda />;
 };
 
 export default ShiftsPage;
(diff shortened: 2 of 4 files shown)


mbhewitt:
## Independent review — LGTM (one minor UX note, no blockers)

Verified the diff (`origin/main...feat/shifts-agenda-flip`) plus the load-bearing surrounding files. The flip is correct and free of redirect loops, dead references, and gating regressions.

### Findings

**Minor (pre-existing, not introduced here) — transient browse-only flash for signed-in users**
`ShiftsAgenda.tsx` reads `shiftboardId` from `SessionContext`, whose initial value is `0` (`context.tsx:37`) and only hydrates from sessionStorage in a `useEffect` (`context.tsx:65-77`). `Schedule` treats `shiftboardId <= 0` as logged-out browse-only (`Schedule.tsx:91`). So on a hard load of `/shifts`, a signed-in volunteer briefly renders the browse-only agenda (no "You're signed up" markers / My-Shifts toggle) until the hydrate effect fires and the context re-render swaps in their real id. It resolves on its own within a tick — no empty state, no infinite loop (hydrate effect has `[]` deps; the persist effect keys on `sessionState` which is stable). This flash is inherent to the SessionContext pattern, not caused by this PR, and the old `/volunteers/[id]/schedule` route dodged it by getting the id from the URL param. Calling it out only so it's a known quantity; not a blocker.

### Checked and clean
- **No redirect loop.** `schedule/page.tsx` now `redirect("/shifts")`; `/shifts` never redirects back. Off-playa an unauth user hitting `/shifts` goes to `/sign-in?returnTo=/shifts` (middleware), not back to the old route.
- **Middleware gating unchanged and correct.** Off-playa `/shifts` requires the `census-session` cookie; on-playa it's allowlisted (`middleware.ts:74`) so walk-ups (id 0) get browse-only. No path where a signed-in user is forced browse-only server-side, and no path where an unauth user hits an error (page → redirect, not 401).
- **No dead code / dangling refs.** `CalendarMonthIcon` import removed from Header and no longer referenced there (still used inside `Schedule.tsx`, a separate file). `shiftboardId` is still used in Header (Account link + BS redirect), so no unused-var. No remaining links to `/volunteers/[id]/schedule` anywhere in `client/src` or `client/e2e`. Removed `AuthGate`/`ACCOUNT_TYPE_AUTHENTICATED` imports are cleanly gone with the old page body.
- **`<Shifts>` table (revert insurance) is fine** — imported by nothing now, not double-rendered, not touched.
- **`Schedule` already handles both modes** (id 0 → open-list only; id > 0 → personal agenda + eligibility), so feeding it the session id is the right minimal wiring.

PR #547, opened 2026-07-21
copy: Help v2 — multistep login, playa/home icons, hyperlinks, VIP rename

mbhewitt:
**Help v2** — Chipper's batch of Help-page updates from the 2026-07-20 review. Copy + in-page navigation only; no app behavior/logic/data touched.

### What changed
- **Multistep login + redirect explainer** — *Getting started* and *Trainings* are now numbered steps that spell out the Burning Man login redirect: *you'll pop to the BM login (the same one Hive uses), then get sent right back to Census — this is normal, you're not being logged out.* Directly targets the Hive same-login confusion volunteers hit at the end of a course.
- **🏜️ on playa / 🏠 from home icons** — steps that depend on location are marked, and the two new-volunteer account paths (from-home vs at-the-lab) are separated.
- **App link** — `volunteers.census.burningman.org` surfaced in *Getting started* as the shareable "this is where you volunteer" URL.
- **Section rename** — *Your checklist* → *Your Volunteer Information Page (VIP)*, with "(labeled **Account** in the menu)" so it's still findable. (The app-wide Account→VIP rename is still an open question for Captain — this only touches Help copy.)
- **Hyperlinks** —
  - Out-links (Shifts, Contact, the app) open in a **new tab** so readers keep their place in the Help.
  - **In-page jump links** auto-open + scroll the target accordion via URL hash (also works when arriving from another page's link, e.g. `/help#trainings`).
- **Acronym** — SAP expanded to *Setup Access Pass* on first use.

### Verification
- `npm run build` clean; `/help` compiles static.
- Playwright: deep-link `/help#vip` opens the VIP accordion; clicking an in-page `#trainings` link opens Trainings. ✅

### Deferred / still open (called out on purpose)
- **Screenshots** (pink Sign-in button / checklist / shift card) — held for a follow-up once copy is locked, per Chipper's own maintenance concern (images go stale + would move if the copy pass reshuffles sections).
- **Tablet-fallback wording** — kept the generic "tell a shift lead / Census leadership" line; needs Chipper's real current procedure.
- **Chipper's copy pass** — copy is hers; markups welcome.

Holding deploy for review. Copy sits within Chipper's authority; this is the "something final to look at."

### DIFF
--- a/client/src/app/help/Help.tsx
+++ b/client/src/app/help/Help.tsx
@@ -15,9 +15,28 @@ import {
   Typography,
 } from "@mui/material";
 
+import { useEffect, useState } from "react";
 import type { ReactNode } from "react";
 
 import { Hero } from "@/components/layout/Hero";
+
+// App URL — shown as shareable text and used for the "this is where you
+// volunteer" link.
+const APP_URL = "https://volunteers.census.burningman.org";
+
+// Links out to another page open in a new tab so readers keep their place in
+// the Help.
+const PageLink = ({ href, children }: { href: string; children: ReactNode }) => (
+  <a href={href} target="_blank" rel="noopener noreferrer">
+    {children}
+  </a>
+);
+
+// In-page jump to another Help section. A plain hash anchor fires hashchange,
+// which the page listens for to open + scroll to the target accordion.
+const Jump = ({ to, children }: { to: string; children: ReactNode }) => (
+  <a href={`#${to}`}>{children}</a>
+);
 
 // Bulleted list that renders naturally inside an accordion.
 const Bullets = ({ items }: { items: ReactNode[] }) => (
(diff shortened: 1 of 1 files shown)


mbhewitt:
Independent review of the hashchange / accordion logic (correctness only — build + the two happy-path Playwright checks already pass). Two real issues:

**[medium] Clicking a jump link to a section you're already anchored at silently does nothing.** `Jump` renders `<a href="#vip">`. If `window.location.hash` is already `#vip`, clicking another `#vip` link does not change the hash, so the browser fires **no** `hashchange` event and `openFromHash` never runs — the accordion doesn't (re)open and nothing scrolls. This is reachable: multiple sections link to `#vip`, and a user can collapse the VIP accordion by clicking its summary while the URL still says `#vip`; a subsequent VIP jump then appears dead. (The Playwright checks each used a fresh/changing hash, so they wouldn't catch it.) Fix: give `Jump` an `onClick` that calls the same open+scroll logic directly instead of relying solely on the `hashchange` event, e.g. hoist `openSection(id)` and call it from both the effect and the click handler.

**[low] scrollIntoView fires before the accordion has expanded.** In `openFromHash`, `setOpen(...)` is queued but `scrollIntoView` runs synchronously in the same tick, before React re-renders and the target accordion grows to full height. For sections low on the page the final scroll position can land short because layout expands after the scroll starts. A `requestAnimationFrame` (or `setTimeout(…, 0)`) around the `scrollIntoView` would let the expansion land first.

Everything else checks out: effect deps `[]` are correct (no stale closure — it reads live `window.location.hash` and uses the functional `setOpen` updater), listener cleanup is correct, SSR-safe (client component, window only touched in effect), toggle Set logic is correct, MUI `expanded`/`onChange` and `List component=\"ol\"` usage is right.

Separately, not a bug but worth a conscious call: `PageLink` forces `target="_blank"` on the **internal** links too (`/shifts`, `/contact`), so in-app navigation opens a new browser tab and drops SPA routing. If that's intended for the internal ones, fine; if not, only the external links (profiles/discord/APP_URL) need `_blank`.

mbhewitt:
Fixed both in `5272172`:

- **[medium] same-hash jump** — `Jump` now calls `openSection` directly on click (with `history.replaceState`) instead of relying on `hashchange`, so a `#vip` link works even when the hash is already `#vip` and the accordion was collapsed via its summary. Verified via Playwright: land on `/help#vip` → collapse VIP → click a `#vip` jump elsewhere → VIP reopens.
- **[low] scroll race** — `scrollIntoView` now runs inside `requestAnimationFrame` so the accordion expands first.

On the third note (internal out-links opening in a new tab): keeping `target=_blank` on `/shifts` and `/contact` on purpose — Chipper explicitly asked for new-tab so readers don't lose their place in the Help. The SPA-routing tradeoff is acceptable for that.

PR #553, opened 2026-07-22
feat(labels): super-admin Avery 5523 volunteer schedule+passcode labels PDF

mbhewitt:
## What

Port of the legacy `vold.census.burningman.org/census.php?function=userPrint` label sheet to the app. New super-admin endpoint:

```
GET /api/labels        # the PDF
GET /api/labels?X=X    # passcodes masked as xxxx (legacy parity)
```

One 2"x4" label per volunteer on Avery 5523 stock (10-up on Letter): red ZapfDingbats dot when Behavioral Standards is unsigned (role 1000012), green star when RS test-out is passed (role 1000011), playa name + italic last name, reversed `terceS susneC:` passcode, on-playa schedule (shifts on/after PreWed from `op_dates`), "Printed at …" / "New On Playa Volunteer" footer.

## Fidelity to legacy

Geometry + font metrics were measured from a PDF generated by the live legacy page (font sizes, baselines, 144pt row pitch, column offsets), and behaviors replicated exactly:
- ≤7 shifts → 9pt rows; 8–9 shifts → all rows at 8pt; 10+ → capped at 7 rows + `****You have N more shifts not listed.****` (N = total−8, legacy's own arithmetic)
- sort: volunteers without a last name first, then by last name + playa name
- position names strip "Census "/"Pre Event "/" 5 SAP", 21-char cap

One deliberate deviation: an extreme-length name is trimmed to stay clear of the passcode instead of wrapping (legacy wrap broke the label's row layout).

## Testing

- Generated locally against the local prod-copy DB; visually compared page-by-page with the legacy 2025/2026 PDFs (normal, dense, overflow, shift-less, and masked labels all verified)
- 401 without session, 403 semantics via `withSuperAdmin` (same guard as SAP endpoints)
- `tsc --noEmit` clean


mbhewitt:
**Verdict: solid port — auth, SQL, nulls, pagination, and pdf-lib usage are all clean, but there is one real fidelity bug (overflow row cap off-by-one vs legacy) that should be fixed before merge, plus one undocumented scope deviation.**

### Findings

**1. Overflow cap is 7 rows; legacy prints 8 — one shift vanishes uncounted (medium)**
`client/src/pages/api/labels/index.ts:199` slices to 7 rows for 10+ shifts, and `:213` prints `total - 8` more. In the legacy second pass (`censusprintsched.php` ~line 122), the `$toomany` check runs *before* the `$t++` increment, so shifts 1–8 are appended and skipping starts at the 9th — legacy shows **8** rows, and its `$aa = total - 8` message is therefore *accurate*, not buggy. The PR body's "capped at 7 rows + N = total−8, legacy's own arithmetic" mischaracterizes legacy: as written, one shift per overflow label is neither printed nor counted (12 shifts → 7 shown, "4 more" claimed, 5 actually hidden). Fix: `slice(0, 8)` — the existing `- 8` message then becomes correct too. (The ≥10 trigger threshold itself matches legacy `aa>1`; 8–9 all-dense at 8pt also matches.)

**2. Build-week shifts dropped — undocumented deviation (low-medium)**
Legacy buckets shifts between PreTrain and PreWed into `$p` and, whenever a volunteer has any such shift, prints them (P then T, `$outD=$idout.$idshiftP...`). The new query (`index.ts:107-109`) hard-filters `d.date >= PreWed` for everyone, so a volunteer with PreSun–PreTue build shifts loses those rows, and the comment at `index.ts:92-93` states legacy behavior incorrectly. Probably the *right* call for on-playa printing (those shifts are past by print time), but the PR body claims "behaviors replicated exactly" — either port it or document it as a deliberate deviation like the name-trim one.

### Verified clean
- **SQL injection**: none — both queries fully parameterized; the `?X=X` mask param never touches SQL.
- **SQL vs schema**: joins/columns all check out; `GROUP BY` is ONLY_FULL_GROUP_BY-safe; role filter correctly in the LEFT JOIN `ON` (roleless volunteers keep their labels); PreWed subquery is deterministic (`datename` has a UNIQUE key); `ORDER BY st.start_time` is a varchar but stored as `YYYY-MM-DD HH:MM`, so lexical = chronological.
- **pdf-lib**: I ran a live check against the installed pdf-lib — ZapfDingbats `embedFont` + `widthOfTextAtSize`/`drawText`/`save` of `●` and `★` all succeed (both glyphs are in pdf-lib's ZapfDingbats encoding). The `centered()` closure correctly follows the reassigned `page`.
- **Nulls**: `playa_name`/`world_name`/`passcode`/`start_time_text`/`end_time_text` all guarded; null passcode → `"0000"` matches legacy `str_pad(null,4,"0")`.
- **Pagination**: exactly 10 volunteers → 1 page; 11 → second page (`i>0 && slot===0`); slot math fills left-right, top-down like legacy's 2-col rows.
- **Auth**: `withSuperAdmin` → HMAC-verified session + server-side `op_volunteer_roles` super-admin check — same guard as the SAP endpoints, as the PR body claims. Appropriate, since this PDF exposes every volunteer's passcode.
- **Role IDs**: 1000011 = "Passed online RS-test-out", 1000012 = "Signed Behavioral Standards" — both confirmed against `op_roles`.
- **Legacy parity spot-checks**: `positionText` strip order + 21-char cap, reversed `terceS susneC:` label, `padEnd(4,"0")`, sort key `("A"|"Z"+last)+playa`, ordinal/12-hour math — all match.

### Minor notes
- `index.ts:87` `shiftboard_id > 1` (vs legacy `> 0`) excludes the `Admin` stub row — sensible; legacy's `shiftboard_pinfo` join excluded it implicitly.
- Legacy's `$last` ("Printed at") leaked across volunteers (never reset) — the new per-volunteer `shifts.length > 0` logic quietly fixes that legacy bug; footer text will differ from old PDFs for some shift-less volunteers, for the better. Note the timestamp uses server TZ — in a UTC container the printed time won't be playa-local.
- Last name derives from `world_name`'s final token (`index.ts:65`), vs legacy `shiftboard_pinfo.last` — multi-word surnames ("van Dyke") sort/print under "Dyke". Unavoidable given `op_volunteers` has no split name field.
- Pre-existing, not this PR: `st.canceled` (`index.ts:110`) isn't in `database/scheduler_schema.sql` even though live DBs and five existing endpoints use it — the checked-in schema is stale.

— Independent review (Claude subagent)

mbhewitt:
Review findings addressed:

**1. Overflow off-by-one — confirmed and fixed** in e8e9258: cap is now 8 rows, so `total − 8` counts exactly the hidden shifts (verified against local data: 12 shifts → 8 rows + "4 more"; 10 → 8 + "2 more"). Re-rendered and the 8-row + note + footer stack still fits the 2in label.

**2. Build-week (PreTrain→PreWed) shifts — not a real behavior difference.** In `DbyY()`, `PreTrain` is **PreSat 17:00**, i.e. *after* PreWed midnight, not before it. The elseif chain tests `> PreWed` first, so any shift that could satisfy `> PreTrain` or `> Mon 12:00` was already caught by the first branch — the $p and $x buckets are dead code with the current date table. The `d.date >= PreWed` filter reproduces the only reachable legacy path. (Confirmed against the live legacy output: its labels contain no pre-PreWed shifts.)

PR #554, opened 2026-07-22
feat(mail): authenticated SMTP for the on-playa build (Gmail as mu@burningman.org)

mbhewitt:
## What

On playa there is no local Exim/SES relay, so the mail worker's queue could drain nowhere — every send hit a dead `127.0.0.1:25` and retried forever. This adds authenticated-submission support to the existing transport:

- `SMTP_SECURE=1` → implicit TLS (`secure: true`) with **full cert verification** (the relaxed `rejectUnauthorized:false` stays only for the localhost relay shape)
- `SMTP_USER` / `SMTP_PASS` → SMTP AUTH; unset = unauthenticated localhost hop, exactly as before (cloud prod unaffected)
- worker boot log now shows the auth mode

The retry queue (`op_email_queue`, backoff capped at 24h, connection errors always transient) already handles intermittent internet — mail queues during outages and flows when the uplink returns.

## Config

Playa `client/.env.production` keys documented in `OPS_Install.md` — Gmail `smtp.gmail.com:465` as `mu@burningman.org` with the legacy app password from `CensusSecret/bm_org_local.php` (same account/credential the PHP `censusmail.php` used), `MAIL_FROM` = legacy `Mu Census VC <censusvolunteercoordinators@burningman.org>`.

## Testing

- Live `verify()` handshake against smtp.gmail.com:465 with the mu credential: OK
- Live send through these exact transport options delivered to Mew's inbox (Gmail 250 OK, From alias honored)
- `npm run test:unit` 26/26, `tsc --noEmit` clean
- Cloud shape regression-guarded: no env set → identical transport options as before


mbhewitt:
**Verdict: approve with one fix requested — the nodemailer options are correct for the two documented shapes and cloud behavior is byte-identical, but the auth/TLS coupling has a misconfiguration hazard worth a one-line change, plus two doc corrections.**

### Medium

1. **Cert verification is keyed off `SMTP_SECURE`, not off "are we sending credentials"** — `client/lib/mail/transport.ts:76`. If `SMTP_USER`/`SMTP_PASS` are set but `SMTP_SECURE=1` is forgotten (e.g. someone tries port 587), the transport does opportunistic STARTTLS with `rejectUnauthorized: false` and submits the real Gmail app password over an unverified TLS session — harvestable by an active MITM. The documented 465 path is safe, but the safe path shouldn't depend on remembering a second env var. One-line fix so auth always implies verification:
   ```ts
   tls: config.smtpUser || config.smtpSecure ? undefined : { rejectUnauthorized: false },
   ```

### Verified clean

2. **Nodemailer options** (nodemailer 8.0.10): `secure: true` on 465 = implicit TLS, correct for Gmail; `tls: undefined` with `secure: true` means default full cert verification, as the comment claims. `auth: undefined` when no user = no AUTH attempted.
3. **Backwards compat**: with no SMTP_* env set, `readConfig()` yields `smtpSecure=false, smtpUser=null` → transport options are `secure: false, auth: undefined, tls: {rejectUnauthorized: false}` — identical to the pre-PR literal. Cloud shape unchanged.
4. **Secret leak**: boot log (`client/lib/mail/index.ts:65`) prints `auth=<user> secure=<bool>` only; `smtpPass` is never logged anywhere in `client/lib/mail/`. Worker tick errors and `last_error` carry SMTP server responses, which don't contain the password. Clean.
5. **Rate limits vs Gmail**: defaults 1/min, 100/day (`index.ts:24-25`) are far under Gmail Workspace's 2000/day. Fine.

### Notable (acceptable, watch after deploy)

6. **535 bad-auth retries forever** — `client/lib/mail/classify.ts:20`: 535 is 5xx but not in `PERMANENT_RESPONSE_CODES`, so a wrong/rotated app password means every row retries until the 24h backoff cap, forever, and nothing ever goes `dead`. This matches the file's deliberate default-to-transient design and the retry rate is low enough that Google lockout risk is minimal — but a bad password is only visible via `last_error`, not as a failure state. Acceptable; check `op_email_queue.last_error` after first playa boot.
7. **Converse misclassification**: Gmail's quota rejection ("550-5.4.5 Daily user sending quota exceeded") is a 550 → marked permanently `dead`, though it's actually transient. Mitigated by the 100/day default staying well under quota; noting as a known ceiling in case `MAIL_RATE_PER_DAY` ever gets raised near Gmail's limit.

### Doc fixes (OPS_Install.md)

8. Column name: the queue column is **`state`**, not `status` (`client/lib/mail/queue.ts:132,178`). The monitoring bullet lists `status`.
9. "Rebuild/recreate the container" — recreate alone suffices: `docker-compose-playa.yaml` loads `client/.env.production` via `env_file` at container start, and SMTP_* are runtime server vars (only `NEXT_PUBLIC_*` bake at build). Not wrong, just makes the playa ops step heavier than needed.
10. Container name `census-app`, the boot-log format, and the "backoff caps at 24h, never gives up on transient" claims all match the code. Verified.

### Minor

- `SMTP_PASS` is read without `.trim()` (`index.ts:20`) — correct in general (passwords may legitimately contain spaces), but combined with #6, trailing whitespace in the env file produces a silent forever-retrying 535. Ops note at most.
- Pre-existing, PR-adjacent: `client/` has no `.dockerignore`, so `COPY . /client` bakes `.env.production` — now including a real Gmail app password — into image layers in addition to the runtime `env_file`. Same exposure class as the secrets already in that file; flagging only because this PR raises the stakes of the file.
- Unit tests pass (7/7) with the extended `MailConfig`.

— Independent review (Claude subagent)

mbhewitt:
Review fix applied in 9455c5d: cert verification is now strict whenever credentials are present (`smtpSecure || smtpUser`), plus both OPS_Install.md corrections (`state` column, recreate-not-rebuild) and a note documenting the endless-535 signature of a bad password. tsc + unit tests clean.

PR #568, opened 2026-07-29
fix(on-playa): pin mysql, drop nginx (app on :80), base-schema migration, capture dnsmasq

mbhewitt:
Folds the local fixes made on the on-playa box during the **2026-07-29 field deploy** into `main`, so `on_playa_code_update.sh` stops re-breaking. All four were live-verified on the box (app on :80 serving, DB stable on 9.4, migrations applied, mail sending).

## Changes

**1. `database/Dockerfile` — pin `mysql:9.4`**
The deploy script runs `docker system prune -fa`, wiping the cached image, so the rebuild re-pulled `mysql:latest` (now 26.7) onto a 9.4 data dir → `Invalid MySQL server upgrade` → `census-database` crash-loop. Data volume was intact; it just needed the matching engine.

**2. `docker-compose-playa.yaml` — drop nginx, app on port 80**
`census-nginx` only served a static redirect-to-`:3000` page plus an unused `:443`. Removed it; the app now publishes `80:3000` (and keeps `3000:3000` so tablets whose home page is `http://192.168.11.11:3000/` keep working). Frees 443. Decided against TLS: the app needs no secure context (no PWA/service-worker/camera; only the clipboard copy-link button), and self-signed on an offline LAN means a per-tablet CA install + an un-removable Android "monitored network" banner for marginal benefit.

**3. `migrations/000_playa_base_columns.sql` — base-schema columns**
`op_shift_times.canceled` and `op_volunteers.arrival_date_id` are referenced by the app (and required by migration 009's `... AFTER arrival_date_id`) but created by no migration — so on a bare OnPlayaData base DB (zero app migrations applied), `/api/shifts`, shift-eligibility, backfill-notify, the mail worker, and migration 009 all failed. Numbered `000` to run first; guarded via `information_schema` so it's a **no-op on cloud/prod** where the columns already exist.

**4. `dnsmasq/dnsmasq.conf` — capture live connectivity config**
The working connectivity fake-out (real IPs for Android connectivity-check + FCM push + NTP) was an uncommitted local mod on the box. Capturing it here so the box can reset to `main` without losing it and future pulls don't conflict.

## Not included
- `client/.env.production` (gitignored, holds DB creds + the Gmail SMTP block) stays box-local by design.

## After merge — one-time box reconciliation
The box has these as uncommitted local mods, which will block `git pull`. Once merged, on the box:
```
cd ~/CensusScheduler
git checkout -- database/Dockerfile docker-compose-playa.yaml dnsmasq/dnsmasq.conf
git pull   # takes main's versions, which now contain all the fixes
```
(`.env.production` is untracked, so it's untouched.)

### DIFF
--- a/database/Dockerfile
+++ b/database/Dockerfile
@@ -1,4 +1,11 @@
-FROM mysql:9.4
+# MySQL version is pinned per-environment via build arg so each deployment
+# matches its own existing data-dir version (MySQL has no in-place downgrade):
+#   playa      = 9.4  (docker-compose-playa.yaml, ephemeral 9.4 data)
+#   cloud/test = 9.7  (docker-compose.yaml, persistent 9.7 volume)
+# Unpinned `latest` once jumped to 26.7 and crash-looped the on-playa DB after a
+# `docker system prune -fa` + rebuild, hence pinning.
+ARG MYSQL_VERSION=9.4
+FROM mysql:${MYSQL_VERSION}
 
 COPY scheduler_schema.sql /docker-entrypoint-initdb.d
 
--- a/database/Dockerfile
+++ b/database/Dockerfile
@@ -1,4 +1,4 @@
-FROM mysql
+FROM mysql:9.4
 
 COPY scheduler_schema.sql /docker-entrypoint-initdb.d
 
--- a/dnsmasq/dnsmasq.conf
+++ b/dnsmasq/dnsmasq.conf
@@ -5,7 +5,17 @@ log-facility=/dev/stdout
 #address=/clients1.google.com/192.168.11.11
 #address=/clients3.google.com/192.168.11.11
 #address=/connectivitycheck.android.com/192.168.11.11
-address=/connectivitycheck.gstatic.com/142.251.46.227
+# Real IPs for the domains Android probes to decide "am I online?" plus FCM
+# push (mtalk), Play services, and NTP time sync — so tablets show connected
+# and keep time. Everything else (address=/#/) is redirected to the local
+# server. These are current-as-of-2026-07 Google IPs; refresh if tablets start
+# reporting "no internet".
+address=/connectivitycheck.gstatic.com/142.250.80.35
+address=/www.google.cn/142.250.189.195
+address=/mtalk.google.com/173.194.65.188
+address=/alt2-mtalk.google.com/192.178.220.188
+address=/play.googleapis.com/216.239.34.223
+address=/2.android.pool.ntp.org/5.161.111.190
 address=/#/192.168.11.11
 no-negcache
 no-resolv
--- a/docker-compose-playa.yaml
+++ b/docker-compose-playa.yaml
@@ -3,6 +3,9 @@ services:
     build:
       context: database
       dockerfile: Dockerfile
+      args:
+        # on-playa DB is 9.4 (ephemeral) — match it
+        MYSQL_VERSION: "9.4"
     container_name: census-database
     image: burning-man/census-database
     security_opt:
--- a/docker-compose-playa.yaml
+++ b/docker-compose-playa.yaml
@@ -26,18 +26,10 @@ services:
       - "192.168.11.11:53:53"
       - "192.168.11.11:53:53/udp"
     restart: always
-  httpd:
-    build:
-      context: httpd
-      dockerfile: Dockerfile
-    container_name: census-nginx
-    image: burning-man/census-nginx
-    security_opt:
-      - apparmor:unconfined
-    ports:
-      - "80:80"
-      - "443:443"
-    restart: always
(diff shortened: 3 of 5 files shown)


mbhewitt:
## Independent review — LGTM-with-nits

Solid, well-commented fix. The migration is correct and genuinely idempotent, and the playa compose/dnsmasq changes are sound. One thing to check before merge, and a few minor notes.

**⚠️ database/Dockerfile is NOT playa-only — it's shared with the cloud/test compose**
- `database/Dockerfile` is the `build.context` for the `database` service in **both** `docker-compose-playa.yaml` and `docker-compose.yaml` (test droplet). So `FROM mysql` → `FROM mysql:9.4` pins the cloud/test DB too, not just playa.
- Pinning is good hygiene, but the test droplet has a **persistent** `census-db-data` volume. If that volume was last initialized under a mysql `latest` newer than 9.4 (9.4 is mid-2025; latest may be 9.5+), the 9.4 container will refuse to start — MySQL has no in-place data-dir *downgrade*. Worth a quick `SELECT VERSION();` on the running test DB before this lands; if it's ≤9.4 you're fine, otherwise pin to that version or plan a dump+reload.
- Not a problem for **prod** (RDS, doesn't use this Dockerfile) or **playa** (its `database` service has no volume — ephemeral, reseeded from the OnPlayaData dump each deploy). The at-risk target is only the test droplet's persistent volume. `mysql:9.4` itself is a valid Docker Hub tag and matches data-dir version 90400. ✅

**migrations/000_playa_base_columns.sql — correct ✅**
- Guarded stored-proc pattern is valid MySQL 8/9 syntax, genuinely idempotent (re-run is a no-op once columns exist), and a true no-op on cloud/prod (the `information_schema` NOT EXISTS guard skips there). `DROP PROCEDURE IF EXISTS` at both ends leaves the DB clean.
- Column types check out: `arrival_date_id BIGINT NULL` matches `op_dates.date_id bigint`; `canceled TINYINT(1) NOT NULL DEFAULT 0` matches the boolean reads (`backfill-notify`: `AND st.canceled = false`, plus `getShiftList`/Schedule). Both columns are confirmed genuinely read by app code, so the migration is justified.
- `DELIMITER` works because migrations are applied through the `mysql` CLI (consistent with 008). ✅

**Note: nothing auto-runs migrations in order**
- Neither `census-deploy.sh` nor `on_playa_code_update.sh` applies `migrations/*.sql` at all (they load the OnPlayaData dump + `blank_pii.sql`). So the "000 runs before 009" guarantee rests entirely on a human applying them via a sorted glob. `000` does sort first lexically, so that holds if done that way — just flagging that the 009 dependency safety is manual, not enforced. Pre-existing condition, not introduced by this PR.

**docker-compose-playa.yaml — fine ✅**
- `"80:3000"` + `"3000:3000"`: valid — two distinct host ports mapping to the same container port, no bind conflict.
- nginx removal is clean: no other live file references `census-nginx` (the only other hit is the `.wt-shift-sheets` worktree, a different branch).
- Minor: the `httpd/` dir (`Dockerfile`, `public-html`, `redirect.conf`) is still in the tree. The comment says "restore from git history," but it wasn't actually deleted — harmless dead dir; could `git rm` it for tidiness.

**dnsmasq/dnsmasq.conf — fine ✅**
- Syntax valid. Specific `address=/domain/ip` entries beat the `address=/#/` wildcard regardless of order in dnsmasq, so no ordering concern with the catch-all below them.
- Only risk is staleness: hardcoded Google IPs and a single Hetzner NTP IP (`5.161.111.190` for `2.android.pool.ntp.org`) will drift. Failure mode is "tablets show no internet" — same as today — and the comment already flags refreshing. Low risk, acceptable for playa.

**Bottom line:** ship it, but confirm the test droplet's live mysql version is ≤9.4 first so the pin doesn't wedge that DB on next rebuild.

mbhewitt:
Addressed the review in cc1f41d..9ce9ba4:

**The one real risk — shared Dockerfile breaking the test droplet — is fixed.** Confirmed the test droplet is on `VERSION() = 9.7.1` with a persistent `census-db-data` volume, so a flat `mysql:9.4` pin would have crash-looped it. Reworked it as a build ARG instead:
- `database/Dockerfile`: `ARG MYSQL_VERSION=9.4` → `FROM mysql:${MYSQL_VERSION}`
- `docker-compose-playa.yaml` builds with `MYSQL_VERSION: "9.4"` (ephemeral playa data)
- `docker-compose.yaml` builds with `MYSQL_VERSION: "9.7"` (test droplet's 9.7 volume)

Each environment now matches its own data-dir version, and both are hardened against future `latest` drift.

On the nits: leaving the `httpd/` dir in the tree (harmless; git history + the dir both make restore trivial) and the hardcoded connectivity IPs (already comment-flagged as drift-prone) — happy to prune `httpd/` in a follow-up if wanted.

PR #569, opened 2026-07-29
fix(saps): accept 2026 SAP header format

mbhewitt:
## Problem
Uploading the 2026 SAP batch on the Send SAPs page reported **0 ingested, 110 unreadable page(s)**.

The 2026 batch PDFs changed the per-page header from:
- 2025: `Setup Access Pass (SAP) 8/16 & Later`
- 2026: `Setup Access Pass 8/26 & Later` (the `(SAP)` literal is gone)

`RE_DATE` in `client/lib/sapPdf.ts` required the `(SAP)` literal, so every page failed the date match and was quarantined.

## Fix
Make `(SAP)` optional in the date regex. Both 2025 and 2026 formats parse; the body copy on every page ("This Setup Access Pass (SAP) is NOT A TICKET…") cannot false-match because the regex still requires `M/D & Later` after it.

## Tests
New `tests/unit/sapPdf.test.ts` exercises the full `splitSapPdf` path with synthetic pdf-lib pages: 2026 format, 2025 format, and a missing-fields page (quarantined, not thrown). `npm run test:unit` green.

After deploy, re-uploading the same batch PDF is safe — ingest is idempotent on ticket_id.

### DIFF
--- a/client/lib/sapPdf.ts
+++ b/client/lib/sapPdf.ts
@@ -10,7 +10,9 @@ import { getDocument } from "pdfjs-dist/legacy/build/pdf.mjs";
 
 import { burnYear } from "./sap";
 
-const RE_DATE = /Setup Access Pass \(SAP\)\s*(\d{1,2})\/(\d{1,2})\s*&\s*Later/i;
+// 2025 header: "Setup Access Pass (SAP) 8/16 & Later"; 2026 dropped "(SAP)".
+const RE_DATE =
+  /Setup Access Pass\s*(?:\(SAP\)\s*)?(\d{1,2})\/(\d{1,2})\s*&\s*Later/i;
 const RE_YEAR = /Access\s*(\d{4})/i;
 const RE_TICKET = /Ticket ID\s*(\d+)/i;
 
--- a/client/tests/unit/sapPdf.test.ts
+++ b/client/tests/unit/sapPdf.test.ts
@@ -1,0 +1,65 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import { PDFDocument, StandardFonts } from "pdf-lib";
+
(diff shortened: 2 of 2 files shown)


mbhewitt:
**Independent review: LGTM.**

Verified against the diff, `client/lib/sapPdf.ts` in full, and the upload endpoint:

- **Regex correctness / false-match risk**: `RE_DATE`'s optional `(?:\(SAP\)\s*)?` cannot be satisfied by the body copy ("This Setup Access Pass (SAP) is NOT A TICKET...", "...is VALID ON OR AFTER...") — `\s*` only spans whitespace, so no bridge from body text to digits is possible. Probed the regex against a full simulated joined page (body-first ordering, tight/no-space concatenation variants): correct match in every case.
- **Backwards compat**: 2025 header "Setup Access Pass (SAP) 8/16 & Later" still parses (covered by test 2).
- **Sibling regexes**: `RE_YEAR` correctly picks 2026 out of "Black Rock City: Access 2026" ("Access Pass 8/26" can't match — "Pass" intervenes); `RE_TICKET` correctly captures `473816590` from the concatenated "Ticket ID 473816590Confirmation Id39GG89B211922061" (`\d+` stops at 'C'). Nothing else in the parser depends on the dropped "(SAP)" literal.
- **Tests**: valid and passing — ran `node --import tsx --test tests/unit/sapPdf.test.ts`, 3/3 pass. The synthetic PDFs include the real body copy, so the false-match case is regression-guarded. Fits the existing `test:unit` runner with no new deps (pdf-lib already a runtime dep).
- **Upload endpoint** (`api/saps/upload`): unchanged and unaffected; quarantine reporting still works via `unparseable`.

No findings. Minimal, correct diff.

PR #571, opened 2026-07-29
feat(saps): name links + behavioral standards & trainings in SAP To-do

mbhewitt:
Per Mew's request on the SAPs page:

1. **Names are links** to the volunteer's page (`/volunteers/{id}/info` — the page the rest of the app labels "View account"). Off-book rows stay plain text (no volunteer page to link).
2. **To-do criteria for the SAP line** now include, per person:
   - required work days (existing day-by-day logic),
   - **Sign Behavioral Standards** (role 1000012 not held),
   - **each uncompleted training** required by their signed-up positions (completion = volunteer holds the training's `op_trainings.role_id`, same rule as the volunteer info endpoint; both sets are queried set-based, no per-person queries).
3. Cell now reads "Missing N item(s)" (was "day(s)"); tooltip lists the items, e.g. `Missing: PreThur; Census Basics training; Sign Behavioral Standards`.

Verified on local prod build against a prod-pull DB: 42 people currently missing behavioral standards, 58 missing a training; screenshot checked. Display-only — assignment is not gated.

### DIFF
--- a/client/src/app/saps/Saps.tsx
+++ b/client/src/app/saps/Saps.tsx
@@ -12,6 +12,7 @@ import {
   Button,
   Chip,
   Container,
+  Link as MuiLink,
   MenuItem,
   Paper,
   Select,
@@ -26,6 +27,7 @@ import {
   Tooltip,
   Typography,
 } from "@mui/material";
+import Link from "next/link";
 import { enqueueSnackbar } from "notistack";
 import { useRef, useState } from "react";
 import useSWR, { useSWRConfig } from "swr";
@@ -275,7 +277,16 @@ export const Saps = () => {
                         spacing={1}
                         alignItems="center"
                       >
-                        <span>{p.name}</span>
+                        {p.shiftboardId != null ? (
+                          <MuiLink
+                            component={Link}
+                            href={`/volunteers/${p.shiftboardId}/info`}
+                          >
+                            {p.name}
+                          </MuiLink>
+                        ) : (
+                          <span>{p.name}</span>
+                        )}
                         {p.kind === "offbook" && (
                           <Chip label="off-book" size="small" />
                         )}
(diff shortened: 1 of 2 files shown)


mbhewitt:
## Independent review

**Verdict: LGTM — no blocking issues.** Verified against the diff, the volunteer info endpoint, and the local DB.

**SQL correctness (trainings query)** — sound. The `LEFT JOIN ... (vr.shiftboard_id IS NOT NULL) AS completed` can't produce both a completed=0 and completed=1 row for the same (volunteer, training): the NULL row only exists when *no* active role row matches, and duplicate active role rows collapse under DISTINCT (confirmed empirically: 0 conflicting variants, 0 duplicate active (shiftboard_id, role_id) rows in the local prod-pull). A training with NULL `op_trainings.role_id` would read as permanently missing — but the info endpoint behaves the same way (`if (roleId)` leaves completed=false), and no such rows exist (all 6 active trainings have role_ids). mysql2 returns `completed` as 0/1 number; the truthy `if (r.completed)` handles that correctly.

**Consistency with info endpoint** — matches: same join chain, same soft-delete filters, same loose `remove_role=false` role semantics (deliberately not the strict add_role check), same BS role id (1000012 via shared constant). One inherited quirk, pre-existing in the info endpoint too: neither trainings query filters `st.remove_shift_time`, so a signup whose shift *time* was removed still generates training requirements. Fine for parity; flagging for a future shared fix.

**Saps.tsx branch logic** — correct. Off-book rows always get `missing: []` from the API, so the missing-first branch never misfires for them; no-arrival volunteers with missing BS/trainings now correctly show "Missing N item(s)" instead of "—", and "—" still shows when nothing's missing. Copy change day(s)→item(s) matches the new semantics. One behavioral note: Staff (who bypass SAP requirements on the info page) get the same BS/training to-dos here — pre-existing looseness (required days were never staff-gated either), just more visible now.

**Link usage** — `<MuiLink component={Link} href=...>` is the correct MUI 7 + next/link pattern; `/volunteers/[shiftboardId]/info` route exists; the `shiftboardId != null` guard mirrors the existing `rowKey` convention for off-book rows.

**Performance** — negligible: the trainings query returned 139 rows over 293 active signups locally. Note the endpoint's Promise.all now issues 11 concurrent queries against a connectionLimit=10 pool — the 11th just queues (no deadlock), but this page fully saturates the pool per load; worth remembering given the recurring prod pool wedge.

**Typecheck** — `tsc --noEmit` clean apart from a pre-existing stale `.next` artifact error unrelated to this PR.

Issue #313, opened 2026-05-23
Email: notify VC list when a volunteer self-drops a critical shift


mbhewitt:
Depends on #307 (email infrastructure foundation). Related to #308 (critical-shift self-removal flow).

## Why

> "I like the idea of a drop notification to the vc list." — @mbhewitt 2026-05-23

Per #308: a volunteer can self-remove from a critical position after acknowledging the warning dialog ("This is a critical shift, are you sure? This leaves your fellow volunteers in a lurch. Please contact a volunteer coordinator for options."). The drop *completes* — but the coordinators need to know it happened so they can move fast on refilling.

Without this, the drop is invisible to Rescue / the coordinators until either (a) someone notices the slot is open, or (b) the shift goes uncovered on the day. Pairs with the warning-dialog flow to close the loop.

## Trigger

The `/api/volunteers/[shiftboardId]/shifts` DELETE handler. After confirming the removal succeeded, check whether the position was marked critical (`op_position_type.critical = 1`). If so, enqueue an email to the VC list via #307.

## Content

- **To:** volunteer-coordinators list — probably `censusvolunteercoordinators@burningman.org`. Same recipient as #312; configurable via env.
- **Subject:** `[Census] CRITICAL OPENING: {Position} on {DayName} {Date}`
- **Body** (plain text + html):
  - Header: "A volunteer just dropped a critical position. The slot is now open."
  - Shift details: date, time, position, shift category
  - Volunteer who dropped: `playa_name` and `world_name` (so the coordinator can follow up if needed)
  - Link to the shift's volunteer-management page: `/shifts/{timeId}/volunteers`
  - Note: the volunteer was shown a warning dialog before this drop completed, so they're aware they left a critical gap.

## Acceptance

- Volunteer self-removes from a critical position via the UI (after dismissing the #308 warning): coordinators get the email within ~1 minute (#307 rate limit).
- Volunteer self-removes from a *non*-critical position: no email — only fires when `pt.critical = 1`.
- Admin removes a volunteer from a critical position: also fires (different request shape, same outcome — slot is now empty and needs refilling).

## Out of scope

- Per-coordinator opt-out — coordinators are a small list and this is high-signal.
- SMS / push — email-only for v1.
- A "request removal" workflow (option 3 from #308 design discussion) — rejected in favor of the warning-dialog approach. If we revisit later, this notification still applies.

🤖 Filed by Claude Code on behalf of @mbhewitt + @chippper_ + Rescue (Ann)

PR #572, opened 2026-07-30
Self-remove shifts for regular volunteers + #308 critical-drop warning

### LINKED ISSUE #308: Block volunteers from self-removing from critical positions (Shiftboard parity)
Filed at @chippper_'s request after the 2026-05-22 team meeting. Reporter: Rescue (Ann).

## Why

In Shiftboard, when a position is marked **critical**, the assigned volunteer cannot self-remove — they must contact an admin. This is load-bearing for Rescue's coordination workflow: if a lead/driver drops a critical position silently, the slot stays in-app marked "filled" and the shift goes uncovered.

> "The critical shifts are set in shiftboard so volunteers can't remove themselves from a shift but have to let us know and we remove them. Will this be true in the tablet software as well? It's important that I know if someone needs to drop a critical shift so I can refill it." — Rescue, 2026-05-22

## Current behavior in CensusScheduler

The "Remove shift" menu item on `/volunteers/[id]/info` (formerly `/account`) is gated only on time-window:

```ts
// client/src/app/volunteers/[shiftboardId]/account/VolunteerShifts.tsx
switch (checkInType) {
  case SHIFT_FUTURE:  isVolunteerRemoveAvailable = true; ...
  case SHIFT_DURING:  isVolunteerRemoveAvailable = true; ...
  case SHIFT_PAST:    isVolunteerRemoveAvailable = isAdmin; ...
}
```

No critical-position check. The response shape `IResVolunteerShiftItem` from `/api/volunteers/[id]/shifts` doesn't even include a critical flag — `op_position_type.critical` exists in the DB and is admin-settable on `/shifts/positions/[id]`, but it isn't threaded through to the volunteer-facing UI.

## Design — TBD

**Disabling the Remove button (Shiftboard parity) is rejected** — that's the Shiftboard band-aid, not the right solution. Per @chippper_: needs a better UX.

Three candidate shapes that solve Rescue's underlying need (knowing about openings in time to refill) without dead-ending the volunteer:

1. **"I can't make this shift" flow.** Button present, but for critical positions it routes through a *request-removal* state instead of immediate delete. Slot stays held until an admin acts. Admin notification fires (depends on #307). Admin removes + refills, or negotiates with the volunteer.
2. **Self-remove + alert.** Allow the remove immediately, but fire an admin notification ("CRITICAL OPENING: {position} on {date} just opened — {volunteer} dropped"). Slot reopens for refill at once. Most permissive; least coordination overhead.
3. **"Find a replacement first."** Critical-position remove only completes after the volunteer either (a) names a swap partner from the volunteer list, or (b) admin approves. Most ownership on the volunteer; highest friction.

Shape decision pending @mbhewitt + @chippper_ + Rescue. Once decided, implementation is straightforward:
- Thread `pt.critical` through `/api/volunteers/[id]/shifts` into `IResVolunteerShiftItem`.
- Implement the chosen flow in `VolunteerShifts.tsx` + the matching DELETE handler in `/api/volunteers/[id]/shifts` (belt-and-suspenders: forged DELETE requests for critical positions must still go through the same path).

## Out of scope

- Notification email infrastructure itself (#307) — but options 1 and 2 depend on it.
- "Critical position" tooltips / explanation copy — UX polish.

🤖 Filed by Claude Code on behalf of @chippper_ + Rescue (Ann)

mbhewitt:
Closes #308.

Logged-in regular volunteers can now remove themselves from shifts they signed up for, from both places they actually look:

## What's in here

**`/shifts` (agenda)** — cards marked "✓ You're signed up" get a **Remove** button (future/during shifts only; past mirrors the account page and stays admin-only). The button opens the existing remove-confirmation dialog and the agenda updates in place — no reload, the card flips back to its open state.

**`/shifts/[timeId]/volunteers` (shift detail)** — non-admins now get an **Actions** menu on *their own* roster row with a single "Remove shift" item. Other volunteers' rows are untouched, and the admin view (Admin review / Admin actions columns) is unchanged.

**#308 critical-drop warning (warn-then-allow)** — both remove dialogs now show a warning when the position is `critical=1` and require an "I understand this leaves a critical position unfilled" checkbox before the Remove button enables. Not a hard block, per the standing decision. The CVC-list email (#313) still fires server-side. This also lights up on the account page's existing remove flow for free.

**Security fix** — `DELETE /api/shifts/[timeId]/volunteers` previously accepted any authenticated user removing *anyone* (only the UI hid it). `shiftVolunteerRemove` now enforces `isOwnerOrAdmin`, covering both DELETE mounts at once. The PATCH check-in path has the same gap — filed separately as a follow-up issue.

**Bug found while testing** — the volunteers DataTable row arrays always carried 6 cells while non-admins only had 4 columns; MUIDataTable maps cells by index and silently drops extras, so the new menu never rendered. Rows are now composed in lockstep with the pushed columns.

## Verification (local, prod-pull DB, headless Chromium)

- Regular user on `/shifts`: Remove on own cards, click doesn't navigate (card is a Link), dialog opens, confirm → card reverts to open, DB row gets `remove_shift=1`
- Critical position: Remove button disabled until the acknowledge checkbox is ticked; non-critical dialog unchanged
- `/shifts/61/volunteers` as regular user: menu only on own row; after remove, roster updates live and "Add this shift" re-enables
- Admin view on the same page: columns and menus identical to before
- Authz: cross-user DELETE with a valid session → **403**; unauthenticated → **401**; self → 200
- `tsc --noEmit` and `npm run build` clean
- e2e: specs 03/04/07/13 fail identically on unmodified `main` (pre-existing local-env issue), no new failures introduced

### DIFF
--- a/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
+++ b/client/src/app/shifts/[timeId]/volunteers/ShiftVolunteers.tsx
@@ -87,6 +87,7 @@ interface IShiftVolunteersProps {
 interface IDialogCurrentState {
   dialogItem: number;
   shift: {
+    critical: boolean;
     positionName: string;
     timePositionId: number;
   };
@@ -123,6 +124,7 @@ export const ShiftVolunteers = ({
   const [dialogCurrent, setDialogCurrent] = useState<IDialogCurrentState>({
     dialogItem: 0,
     shift: {
+      critical: false,
       positionName: "",
       timePositionId: 0,
     },
@@ -182,6 +184,9 @@ export const ShiftVolunteers = ({
             if (dataShiftVolunteersItem) {
               const dataMutate = structuredClone(dataShiftVolunteersItem);
               dataMutate.volunteerList.push({
+                // socket payload has no critical flag; mutate() revalidates
+                // right after, which fills in the real value
+                critical: false,
                 isCheckedIn,
                 notes,
                 playaName,
@@ -395,6 +400,14 @@ export const ShiftVolunteers = ({
     isVolunteerAddAvailable = false;
     isCheckInAvailable = false;
   }
+  // Non-admins may remove THEMSELVES from a future/during shift — mirrors the
+  // account page (past shifts stay admin-only). Canceled shifts intentionally
+  // still allow self-removal (see the comment above). Admins keep the full
+  // remove-anyone menu.
+  const isSelfRemoveAvailable = (rowShiftboardId: number): boolean =>
+    isAuthenticated &&
+    rowShiftboardId === shiftboardId &&
+    checkInType !== SHIFT_PAST;
 
   // prepare datatable positions
   const columnListPositions = [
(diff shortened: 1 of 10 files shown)


mbhewitt:
## Independent review (Claude subagent)

Reviewed `7a97886` vs `main`: both DELETE mounts and the shared guard, MUIDataTable column/cell alignment for all four viewer permutations (admin / signed-up non-admin / non-signed-up / signed-out, desktop + mobile), the shadowed `shiftboardId` in the row maps (closure correctly captures the session id), SWR mutation keys, acknowledge-checkbox reset in all three mount patterns, and canceled/past handling. No HIGH or MED defects. Two LOW findings, **both fixed in 64494e6**:

1. **Stale open-shift counts after an agenda self-remove** — only `/api/volunteers/[id]/shifts` revalidated, so a dropped shift could re-render as "Full" with a count that still included the volunteer until the next focus revalidation. Fixed: the remove dialog now also `mutate("/api/shifts")` on success. Verified in headless Chromium: the dropped shift's card flips to open with the decremented count (3→2 filled) immediately.

2. **Past-shift rule only enforced client-side** — a volunteer could replay the DELETE with their own id and a past `timePositionId`, erasing an already-worked (e.g. no-show) signup from fill reports and prior-year counts. Fixed: `shiftVolunteerRemove` now 403s non-admin removals of shifts that ended >24h ago (the 24h grace means server-vs-playa timezone skew can never block a legitimate during/just-after self-drop). Verified: past-shift self-DELETE → 403, future → 200.

Verified-fine (not findings): row/column index alignment, self-remove gating (own row + not past + authenticated), canceled shifts intentionally removable, both dialogs' mutation keys match their pages' SWR keys, CVC critical-drop email fires server-side independent of the client checkbox, existing admin remove/review flows behavior-identical.

PR #575, opened 2026-07-30
feat(saps): grey rule rework, persisted date choice, unlock+burn flow, ticket serials

mbhewitt:
Per Mew's batch of SAP page requests:

1. **Grey rule** — rows are grey when the person has no CSP-earning shift starting on/before Open Sunday, **or** is Staff — unless someone sets a date (persisted override or an actual assignment), which lifts them back to red/green. External SAP stays grey with `External SAP` text.
2. **Staff dropdown** — the auto option reads just `Staff`, no dayname appended.
3. **Grey ≠ hover** — grey is now `#e0e0e0`; MUI's hover tint (~`#f5f5f5`) reads distinctly on all row colors.
4. **Date choice persists** — new `sap_date_override` column on `op_volunteers` / `op_sap_offbook` (**migration `010_sap_date_override.sql` — apply by hand before deploy**), saved via new `POST /api/saps/date-choice` whenever the dropdown changes. Assign uses it when no fresher choice is made.
5. **Unlock + burn** — clicking the lock on a delivered SAP asks `are you sure` (explains the burn), then re-opens the date dropdown + assign checkbox. Assigning a new SAP burns the old one via the existing assign-endpoint logic (`superseded_by_sap_id` chain); burned SAPs can never be reassigned.
6. **Ticket serial in filenames** — `SAP_<date>_<ticketId>.pdf` for download, re-download, and the email attachment.
7. **Email copy** — subject now includes “Census” (standing rule); body adds: addressed to Matthew Hewitt, works for you, contact Census if Gate issues.

Verified on local prod build + prod-pull DB (migration applied locally): date-choice round-trips and clears; staff without a date all grey; standings 57 grey / 20 red / 12 external / 4 green; screenshot checked. 34 unit tests green.

### DIFF
--- a/client/lib/sapStatus.ts
+++ b/client/lib/sapStatus.ts
@@ -49,8 +49,9 @@ export interface RequiredDay {
 
 // Where a person stands on getting a SAP:
 // - external: getting a SAP from another group (role 2000007) — not ours to earn
-// - not_earning: no SAP-earning shift signed up (and not staff), so currently
-//   not on track to get one
+// - not_earning: not on track to get one by default — no CSP-earning shift
+//   before/on Open Sunday, or Staff. An explicit date (admin override or an
+//   actual assignment) lifts this back to missing/complete.
 // - missing: on track but requirements outstanding
 // - complete: all requirements met
 export type SapStanding = "external" | "not_earning" | "missing" | "complete";
@@ -62,7 +63,19 @@ export interface SapEligibilityInput {
   missingTrainings: string[]; // names of required-but-uncompleted trainings
   totalCsp: number;
   requiredDays: RequiredDay[];
-  hasEligibleShift: boolean; // has at least one CSP-earning shift
+  hasPreOpenShift: boolean; // has a CSP shift starting on/before Open Sunday
+  hasDateOverride: boolean; // admin picked a date or a SAP is assigned/issued
+}
+
+// A SAP is only relevant when the first CSP-earning shift starts on or before
+// Open Sunday — later shifts need no early entry. Dates are YYYY-MM-DD, so
+// string compare is safe. Unknown Open Sunday -> any CSP shift counts.
+export function isPreOpenShift(
+  firstCspShiftDate: string | null,
+  openSunDate: string | null,
+): boolean {
+  if (!firstCspShiftDate) return false;
+  return openSunDate === null || firstCspShiftDate <= openSunDate;
 }
 
 export interface SapEligibility {
@@ -110,7 +123,7 @@ export function evaluateSapEligibility(
 
   const standing: SapStanding = input.hasExternalSap
     ? "external"
-    : !input.hasEligibleShift && !input.isStaff
+    : (input.isStaff || !input.hasPreOpenShift) && !input.hasDateOverride
       ? "not_earning"
       : requirementsMet
         ? "complete"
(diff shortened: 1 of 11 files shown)


mbhewitt:
## Independent review — feat/sap-page-followups

**Verdict: approve.** Unlock/burn flow, grey rework, date-choice persistence, and ticket serials are all correct. Tests pass (ran the branch's sapStatus suite: 5/5). Findings, ranked:

**1. MUI Select out-of-range when the override date has no available passes (minor, real).** `choice = dateChoice[key] ?? p.dateOverride ?? "auto"`, but the menu only contains "auto" + `availableDates` (dates with `status='available'` stock). Reachable two ways: (a) admin pins date X for person A, last X pass gets assigned to person B → A's dropdown value X is absent (MUI warning, blank-looking select); (b) unlocked received row whose override equals the delivered date and no second pass of that date exists. Fix is one line: render a disabled MenuItem for the override date when it's not in `availableDates`.

**2. Same-date reassign after unlock: safe, but consumes a second pass (behavioral note).** The assign endpoint's same-date no-op short-circuit only matches `status='assigned'`, so an unlocked *received* row assigned the same date reserves a **new** pass of that date and burns the old one. If no second pass of that date exists → clean 409, nothing burned, old SAP stays valid — correct failure mode. But an admin who only wants to *re-send* will waste a physical pass by unlocking; re-send already works via email/download without unlocking. Consider a line in the confirm dialog: "To re-send the existing pass, use the download/email buttons instead."

**3. Unlock/burn flow verified correct.** Burn happens only after the replacement is reserved, in one transaction under the per-person named lock; failure rolls back with the received SAP intact. DELETE /assign only touches `status='assigned'` so a received SAP can never be freed — and the UI can't even fire DELETE on an unlocked row (it renders unchecked; clicking assigns). Assign error path keeps the row unlocked for retry; success clears `unlocked` before refresh. No staleness found; unlock resetting on page reload is fine (old SAP untouched until assign).

**4. date-choice endpoint: solid, two nits.** Parameterized SQL (no injection), `withSuperAdmin`, exactly-one-of target validation, and the `affectedRows===0 → 404` check is safe on no-op re-saves (mysql2 default flags include FOUND_ROWS, so matched-but-unchanged rows count). Nits: the regex accepts non-calendar dates (`2026-13-99`) which would surface as an unhandled 500 from MySQL strict mode instead of a 400 — unreachable from the UI (values come from `availableDates`); and email is lowercased but not trimmed, while the offbook insert stores `trim().toLowerCase()` — fine for the UI (sends DB-normalized emails), only raw API callers with padding would get a 404.

**5. isPreOpenShift parity + grey semantics: correct.** Both endpoints look up `OpenSun` from `op_dates` and call the same helper; `dateStrings: true` on the pool guarantees YYYY-MM-DD so the string compare is valid. Missing OpenSun row degrades to "any CSP shift counts" (old behavior) identically in both. `hasDateOverride = assignment || override` propagates for volunteers and off-book.

**6. Volunteer info endpoint unaffected, as intended.** It consumes only `requirementsMet`, which is computed purely from BS/trainings/CSP/required-days — independent of the new `hasPreOpenShift`/`hasDateOverride` inputs. `standing` is unused there.

Also checked: migration 010's unguarded ALTER matches 009's convention; ticket-serial filenames fall back to sapId and ticket_id is digits-only in practice.

🤖 Independent review by Claude Code subagent

PR #576, opened 2026-07-30
feat(saps): red 'SAPs have been uploaded' note next to the upload button

mbhewitt:
Shows `The {year} SAPs have been uploaded` in red next to the upload button whenever the pool has any SAPs for the current burn year. Verified on local prod build (DOM check: text renders, MUI error red).

### DIFF
--- a/client/src/app/saps/Saps.tsx
+++ b/client/src/app/saps/Saps.tsx
@@ -259,7 +259,10 @@ export const Saps = () => {
         </Typography>
 
         {/* upload */}
-        <Box component="section" sx={{ mb: 4 }}>
+        <Box
+          component="section"
+          sx={{ mb: 4, display: "flex", alignItems: "center", gap: 2 }}
+        >
           <input
             ref={fileInputRef}
             type="file"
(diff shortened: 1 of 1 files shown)


mbhewitt:
**Review: LGTM** — quick verdict on the three asks:

- **Null-safety**: `(poolData?.saps.length ?? 0) > 0` is safe — during load `poolData` is `undefined`, the optional chain short-circuits to `undefined`, `?? 0` hides the note. If the pool fetch errors, `poolData` stays undefined and the note just stays hidden (fine for a cosmetic hint). `burnYear` is guarded by the existing `if (!peopleData) return <Loading />` early return.
- **Layout**: the section Box going `display: flex; alignItems: center; gap: 2` is harmless — the file `<input hidden>` keeps its UA `display: none` (hidden flex children don't participate in layout), so it's just Button + note side by side, vertically centered.
- **Condition semantics**: GET /api/saps returns every `op_saps` row for `getCurrentBurnYear()` (all statuses, incl. assigned/received/burned), and returns `saps: []` when burnYear is null — so `length > 0` correctly means "this year's batch has been uploaded", and the "The null SAPs" edge case can't render (null burnYear ⇒ empty pool ⇒ note hidden). Pool and people endpoints share the same `getCurrentBurnYear()`, so the interpolated year matches the pool being counted.

One cosmetic nit, not blocking: the note pops in a beat after the button while the pool fetch resolves. Fine as-is.

